PR #758, opened 2022-04-25
tools/generator-go-sdk: normalizing the casing of Constants returned from the API

### LINKED ISSUE #417: generator-go-sdk: unmarshaling should normalise the casing of Enum's
Enum's are a Type Alias of String - as such we should be able to add a custom Unmarshaler to normalise the casing on Enum Values, which would work around issues in the API where these are sometimes returned in a differing case to what we expect.

Note that this is intentionally only for Unmarshaling and not for Marshaling of Enum Values - which means these'll be consistent on Create (since we specify the values) and Read (since we're normalising the casing on the way back).

~At the same time we should look to add the `Possible{Name}Values` method which we can use for schema validation functions too~ added in #272

tombuildsstuff:
This PR updates the custom unmarshal funcs output by models to support normalizing the casing of constants from whichever casing is returned in the API - to the casing used in the API Definitions. Whilst this may seem like an odd choice at first, this removes a non-trivial amount of logic errors from the API where the Service Team have incorrectly defined the casing in the Swagger vs what's returned from the API - working around the situation where the ARM Team won't let the Service Teams update the casing in the Swagger to match the response coming back from the API.

~Dependent on #757~
Dependent on https://github.com/hashicorp/go-azure-helpers/pull/113

Fixes #417


tombuildsstuff:
Closing in favour of #2280

PR #1309, opened 2022-08-10
Support for SDK base layer

manicminer:
Depends on: ~https://github.com/hashicorp/go-azure-sdk/pull/122~ (merged)
Depends on: ~https://github.com/hashicorp/go-azure-sdk/pull/276~ (merged)


tombuildsstuff:
I've rebased this branch & removed the Data API changes, meaning that the `generator-go-sdk` now contains all of the logic for which base layer should be used (since this is intended as temporary logic), having chatted with @manicminer, we're going to handle this transition in two ways:

* Initially we'll use an explicitly defined list of the Services which should be generated using the new base layer.
* Once we've proven this works as expected, we'll invert this logic - giving us a list of Services which should continue using the old base layer.

Doing so will allow us to first roll out a handful of Services, address any issues we've found - and then gradually switch all existing Services over by updating the list & the usages within `hashicorp/terraform-provider-azurerm`.

tools/generator-go-sdk/generator/templater_clients.go
```diff
@@ -0,0 +1,40 @@
...
+%[3]s
+
+type %[2]s struct {
+	Client  *resourcemanager.Client
+}
+
+func New%[2]sWithBaseURI(api environments.Api) (*%[2]s, error) {
+	client, err := resourcemanager.NewResourceManagerClient(api, %[1]q, defaultApiVersion)
+	if err != nil {
+		return nil, err
```
tombuildsstuff, line 32: do we want to wrap this error to provide context about _which_ client failed?

tools/generator-go-sdk/generator/templater_methods.go
```diff
@@ -0,0 +1,672 @@
...
+		return nil, fmt.Errorf("building arguments for list operation: %+v", err)
+	}
+	requestOptions, err := c.requestOptions(data)
+	if err != nil {
+		return nil, fmt.Errorf("building request config: %+v", err)
+	}
+	argumentsCode := c.argumentsTemplate()
+	unmarshalerCode, err := c.unmarshalerTemplate(data)
+	if err != nil {
+		return nil, fmt.Errorf("building unmarahaler template: %+v", err)
```
tombuildsstuff, line 280:
```suggestion
		return nil, fmt.Errorf("building unmarshaler template: %+v", err)
```

tools/generator-go-sdk/generator/templater_methods.go
```diff
@@ -0,0 +1,672 @@
...
+	return &out, nil
+}
+
+func (c methodsPandoraTemplater) marshalerTemplate(data ServiceGeneratorData) (*string, error) {
+	var output string
+
+	if c.operation.RequestObject != nil {
+		output = fmt.Sprintf(`
+	err = req.Marshal(input)
+	if err != nil {
```
tombuildsstuff, line 518:
```suggestion
	if err = req.Marshal(input); err != nil {
```

tombuildsstuff (approved):
Left a few comments inline but 👍

Issue #403, opened 2021-10-27
importer: retrieve and diff existing API Definitions


tombuildsstuff:
Today the importer parses the Swagger files out and then generates the API Definitions - whilst this works it means we don't consider existing fields and if they've been renamed (which happens).

As such during the import of the Swagger data we should take an optional `--data-api-endpoint` parameter (to the Data API Endpoint, locally that's `http://localhost:5000`) and connect to that, pull down the existing service (if available) and then diff these.

This enables us to:

1. Detect breaking changes (which we can flag in a Github Action/comment at import-time in the future)
2. Handle when a Field has been intentionally renamed in the API Definitions to be clearer (we can determine this from the JsonName being the same but the FieldName being different, for fields)
3. Handle when a Constant or it's Keys (but not the Constant Value) has been intentionally renamed in the API Definitions to be clearer (we can determine this from the Constant Value being the same)
4. Handle when a Model has been intentionally renamed in the API Definitions to be clearer (we can determine this from the JsonName of the Type being the same)

Ultimately this should enable us to workaround breaking changes within the Swagger by normalising these at import time - it's not flawless and there's more to add over time - but that should be a good start (and allow us to rename things to be clearer where they make sense - since in time we'll be using the FieldName's for Terraform resources etc).

Since this is optional it's possible to bypass this if required - such as for checking that we can import the existing data in unit tests.

PR #1637, opened 2022-10-03
tooling: running the `importer-rest-api-specs` tool connected to the Data API

tombuildsstuff:
This enables loading existing data from the Data API and diffing it against the imported data.

Initially this is only focused on the Terraform Tests, but will be expanded to other items in the future

Fixes https://github.com/hashicorp/pandora/issues/403


tombuildsstuff:
Closing this one out since it's been sitting indefinitely/will be reworked if needed.

PR #1675, opened 2022-10-10
config: add appplatform@2022-09-01-preview

ms-henglu:


tombuildsstuff (changes requested):
hey @ms-henglu

I've taken a look through and if we can fix the `name` here then this should otherwise be good to go 👍 

Thanks!

katbyte (approved):
LGTM 🧮

PR #1681, opened 2022-10-11
refactor: cleaning up some TODO's

tombuildsstuff:
This PR threads through the Provider Prefix from the top-level of the `generator-terraform` project, and also switches to using the `testhelpers` package where possible, rather than duplicating the test functions


jackofallops (approved):

PR #1685, opened 2022-10-12
config: Update api version for new resource `Network Function`

ms-zhenhua:
[Swagger](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/networkfunction/resource-manager/Microsoft.NetworkFunction/stable/2022-11-01)
[Document](https://learn.microsoft.com/en-us/azure/expressroute/how-to-configure-traffic-collector)

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -221,7 +221,7 @@ service "netapp" {
 }
 service "networkfunction" {
   name      = "NetworkFunction"
-  available = ["2022-11-01"]
+  available = ["2022-08-01", "2022-11-01"]
 }
 service "nginx" {
   name      = "Nginx"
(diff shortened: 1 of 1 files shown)


config/resource-manager.hcl
```diff
@@ -221,7 +221,7 @@ service "netapp" {
 }
 service "networkfunction" {
   name      = "NetworkFunction"
-  available = ["2022-08-01"]
+  available = ["2022-11-01"]
```
katbyte, line 224: instead of changing this in place we should be adding the new version
```
available = ["2022-08-01", "2022-11-01"]
```

katbyte (changes requested):

config/resource-manager.hcl
ms-zhenhua: Hi @katbyte, thank you for your reviewing. I have updated the configuration. Kindly help review it again. Thanks.

katbyte (approved):
Thanks @ms-zhenhua - LGTM 🌵

katbyte:
there is a swagger resource id inconsistency that will need to be fixed before this can be merged: #1700 - i'll reopen this shortly

PR #1757, opened 2022-10-26
add `postgresql@2022-03-08-preview`

ziyeqf:


config/resource-manager.hcl
```diff
@@ -261,7 +261,7 @@ service "portal" {
 }
 service "postgresql" {
   name      = "PostgreSql"
-  available = ["2017-12-01", "2018-06-01", "2020-01-01", "2021-06-01"]
+  available = ["2017-12-01", "2018-06-01", "2020-01-01", "2021-06-01", "2022-03-08-preview"]
```
tombuildsstuff, line 264: can we consolidate the versions being used here at all?

ziyeqf: I think we can not, because some features only exist in a certain version..

ziyeqf: It does too many versions here, but I think it is not able to be consolidated to less. Can we merge this PR cause the newer API version is needed to fix https://github.com/hashicorp/terraform-provider-azurerm/issues/18838

PR #1765, opened 2022-10-27
tools/importer-rest-api-specs: surfacing both Single & Multiple `Zones`

tombuildsstuff:
This PR surfaces both the singular `Zone` and plural `Zones` types in both the Data API and the Go SDK - so that this can be used in the Terraform Generator. In addition this now detects both singular and plural Zone(s) in the Schema and switches the types out - so these CustomTypes can be used as needed.

This PR also fixes a couple of bugs where these weren't fully threaded through, causing `Zones` to be output as an empty type by the Data API once the Importer detected these.


tools/importer-rest-api-specs/components/schema/processors/field_name_rename_boolean.go
```diff
@@ -0,0 +1,48 @@
...
+		// change `allow_X` prefix -> `X_enabled`
+		if strings.HasPrefix(strings.ToLower(fieldName), "allow") {
+			updated := fmt.Sprintf("%sEnabled", fieldName[5:])
+			updatedFieldName = &updated
+		}
+		// change `allowed_X` prefix -> `X_enabled`
+		if strings.HasPrefix(strings.ToLower(fieldName), "allowed") {
+			updated := fmt.Sprintf("%sEnabled", fieldName[7:])
+			updatedFieldName = &updated
+		}
```
jackofallops, line 44: Should these not all be returns, since we'd want to exit processing on match? (and process `allow` / `allowed` the other way around for longest prefix matching)

jackofallops (approved):
Thanks @tombuildsstuff - one question (I feel I may have asked before?) otherwise LGTM 👍

tools/importer-rest-api-specs/components/schema/processors/field_name_rename_boolean.go
tombuildsstuff: Good shout, will fix this in the next PR (since it's based atop this one)

PR #1766, opened 2022-10-27
tools/generator-terraform: fixing mappings for Zones/Constants

tombuildsstuff:
This PR fixes the Mappings for both Zones and Constants


stephybun (approved):
LGTM 🎈

PR #1781, opened 2022-11-02
`importer-rest-api-specs` and `generator-terraform`: various fixes

tombuildsstuff:
This PR includes a number of fixes to both `importer-rest-api-specs` and `generator-terraform`:

1. Ensures that we output the Mappings in the Terraform Generator for an Update function where the SDK Model is shared between a Create and Update
2. `tools/importer-rest-api-specs`: outputting the referenced model name for `ModelToModel` mappings rather than computing it (which maybe wrong if this has been renamed)
3. `tools/importer-rest-api-specs`: outputs the value `acctest` in the Test HCL for fields named `dns_prefix` (there's several of these, and they're prefixes, so this is fine)
4. `tools/importer-rest-api-specs`: handling an optional field containing a required field (such that the block is ultimately required)


stephybun (approved):
LGTM 🦆

PR #1782, opened 2022-11-02
New Resource: generating Kubernetes Fleet Manager

tombuildsstuff:
This PR introduces support for Kubernetes Fleet Manager, which is in Preview.

There's an upstream issue in the Swagger, which is fixed in this PR: https://github.com/Azure/azure-rest-api-specs/pull/21394

To workaround that in the interim this PR includes an override to allow us to generate this resource.


stephybun (approved):
LGTM 🐎

Issue #1223, opened 2022-07-28
Data Issue: Network


tombuildsstuff:
```diff
diff --git a/data/Pandora.Definitions.ResourceManager/Network/v2021_08_01/ExpressRouteCircuitConnections/ResourceId-PeeringId.cs b/data/Pandora.Definitions.ResourceManager/Network/v2021_08_01/ExpressRouteCircuitConnections/ResourceId-PeeringId.cs
index 543e5fe2..e989fe9a 100644
--- a/data/Pandora.Definitions.ResourceManager/Network/v2021_08_01/ExpressRouteCircuitConnections/ResourceId-PeeringId.cs
+++ b/data/Pandora.Definitions.ResourceManager/Network/v2021_08_01/ExpressRouteCircuitConnections/ResourceId-PeeringId.cs
@@ -12,7 +12,7 @@ internal class PeeringId : ResourceID
 {
     public string? CommonAlias => null;

-    public string ID => "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualRouters/{virtualRouterName}/peerings/{peeringName}";
+    public string ID => "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/expressRouteCircuits/{circuitName}/peerings/{peeringName}";

     public List<ResourceIDSegment> Segments => new List<ResourceIDSegment>
     {
@@ -22,8 +22,8 @@ internal class PeeringId : ResourceID
         ResourceIDSegment.ResourceGroup("resourceGroupName"),
         ResourceIDSegment.Static("staticProviders", "providers"),
         ResourceIDSegment.ResourceProvider("staticMicrosoftNetwork", "Microsoft.Network"),
-        ResourceIDSegment.Static("staticVirtualRouters", "virtualRouters"),
-        ResourceIDSegment.UserSpecified("virtualRouterName"),
+        ResourceIDSegment.Static("staticExpressRouteCircuits", "expressRouteCircuits"),
+        ResourceIDSegment.UserSpecified("circuitName"),
         ResourceIDSegment.Static("staticPeerings", "peerings"),
         ResourceIDSegment.UserSpecified("peeringName"),
     };
[...]

PR #1785, opened 2022-11-03
tools/importer-rest-api-specs: resolve common id comparison issue for case sensitive static segment in url of `Network` resource

ms-zhenhua:
fix https://github.com/hashicorp/pandora/issues/1223


stephybun:
Thanks for this PR @ms-zhenhua and apologies for taking some time to circle back to this one. 

It appears this was fixed in #2030 so I'm going to close this one for now.

PR #1794, opened 2022-11-04
`kubernetes_fleet_manager` - outputting that this is in Preview

tombuildsstuff:
This PR adds a disclaimer that the Kubernetes Fleet Manager resource is in Preview - and adds support for multi-line descriptions.

Whilst longer-term I have a feeling we'll want to expose Preview information in a different manner, this should be fine in the interim.


catriona-m (approved):
LGTM!

PR #1798, opened 2022-11-07
tools/generator-terraform: outputting the resource Display Name as the description

tombuildsstuff:
Whilst we could output the regular description here this isn't necessarily valid as YAML FrontMatter - which causes this linting error:

`* website/docs/r/kubernetes_fleet_manager.html.markdown: error checking file frontmatter: error parsing YAML frontmatter: yaml: line 9: could not find expected ':'`

As such this commit changes to using a pre-formatted string for the frontmatter description which whilst isn't perfect (grammatically) works well enough for now

Fixes https://github.com/hashicorp/terraform-provider-azurerm/actions/runs/3395699863/jobs/5645911212


stephybun (approved):
LGTM ☁️

PR #1799, opened 2022-11-07
Data: regenerating based on cabe862a1e49f3c9c243d1361e2907ce1acafb9a

hc-github-team-tf-azure:
This PR is automatically generated based on the commit cabe862a1e49f3c9c243d1361e2907ce1acafb9a

PR #1800, opened 2022-11-07
tools/importer-rest-api-specs: fixing an issue where the Resource Group naming is inconsistent

tombuildsstuff:
This allows incorrectly duplicate segments (such as `sourceResourceGroupName` rather than `resourceGroupName`) to be normalised to `resourceGroupName` providing it contains `resourcegroup` and is preceded by `resourceGroups`


tombuildsstuff:
Fixes #1799

PR #1829, opened 2022-11-16
tools/version-bumper: auto-onboarding new versions of existing APIs

tombuildsstuff:
This PR enables the auto-onboarding of new API Versions for Services which have been imported into Pandora.

This means that should a new API Version become available, and we've not explicitly marked it as that it should be being ignored, that a PR will be sent to add the new API Version to the config when the Swagger is updated.


stephybun (approved):
LGTM 🐦

PR #1833, opened 2022-11-16
tools/importer-rest-api-specs: handling discriminators that inherit from other discriminators

### LINKED ISSUE #1832: Discriminators which inherit from other discriminators aren't pulled out
When a Discriminated Type inherits from another discriminated type, it should be pulled out, but it's not being today, which means that the following Discriminators aren't identified for Media Services:

https://github.com/Azure/azure-rest-api-specs/blob/fa7609844bc20b126037dfb180ef7155c2174f7b/specification/mediaservices/resource-manager/Microsoft.Media/Encoding/preview/2019-05-01-preview/Encoding.json#L1228-L1246

https://github.com/Azure/azure-rest-api-specs/blob/fa7609844bc20b126037dfb180ef7155c2174f7b/specification/mediaservices/resource-manager/Microsoft.Media/Encoding/preview/2019-05-01-preview/Encoding.json#L1247-L1262

tombuildsstuff:
Fixes #1832

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -398,22 +398,48 @@ func (d *SwaggerDefinition) modelDetailsFromObject(modelName string, input spec.
 
 		// so we need to find the parent details
 		for _, parentRaw := range input.AllOf {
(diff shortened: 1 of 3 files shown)


stephybun (approved):
LGTM 🐱 🐶

PR #1877, opened 2022-11-28
Swagger Data Workaround for `recoveryservicessiterecovery`

tombuildsstuff:
This PR adds a Swagger Data Workaround for the service `RecoveryServicesSiteRecovery`.

This PR also refactors the existing Swagger Data Workarounds to use structs for each workaround, which allows for these to be added/removed as required.


tools/importer-rest-api-specs/components/parser/load_and_parse.go
```diff
@@ -2,6 +2,7 @@ package parser
 
 import (
 	"fmt"
+	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/components/parser/dataworkarounds"
```
jackofallops, line 5: goimports?

jackofallops (approved):
One goimports comment, otherwise LGTM 👍

tools/importer-rest-api-specs/components/parser/load_and_parse.go
tombuildsstuff: good shout, ta

PR #1882, opened 2022-11-29
tools/generator-go-sdk: reverting the `optional object` logic for now

tombuildsstuff:
Unfortunately there's a number of APIs this causes issues with (DataBricks, Media, Mixed Reality and Notification Hubs) as such we'll want to revert this for now, raise API issues, then look to re-introduce this


jackofallops (approved):
LGTM 🎸

PR #1984, opened 2023-01-09
Data: regenerating based on fc8a116efcdf8ec9cf12a4d28f2aab52cf4657a6

hc-github-team-tf-azure:
This PR is automatically generated based on the commit fc8a116efcdf8ec9cf12a4d28f2aab52cf4657a6

PR #1985, opened 2023-01-09
tools/importer-rest-api-specs: handling the responses having multiple response objects

tombuildsstuff:
When the operation has multiple response objects, we pick the lowest status code where possible.

As a general rule this should be a 200/201, but this works around operations having different response objects for different operations, since we only parse the HTTP Response when we're expecting a 200.

When a single response object is returned we parse the status code regardless, so this change is limited to operations with multiple status codes.

Fixes the issues seen in #1984


jackofallops (approved):
LGTM 👍

PR #1995, opened 2023-01-16
feat: create authorization service version 2020-10-01

josh-barker:
Hi,

This PR is to create a newer version of the authorization api so that I can make a resource for Role Eligibility Schedule Requests for PIM on Azure Resources.

Would you recommend to use the latest version, i.e. `2022-04-01-preview` or `2020-10-01` as it's stable?

Cheers

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -32,6 +32,10 @@ service "attestation" {
   name      = "Attestation"
   available = ["2020-10-01"]
 }
+service "authorization" {
+  name      = "Authorization"
+  available = ["2020-10-01"]
+}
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
> Would you recommend to use the latest version, i.e. 2022-04-01-preview or 2020-10-01 as it's stable?

The latest stable should be fine here, I've checked and whilst there's a newer stable API version available (`2022-04-01`) it doesn't include this functionality, so this appears to be the latest stable API version supporting this feature

👍

tombuildsstuff (approved):
LGTM - thanks for this @josh-barker

PR #2007, opened 2023-01-19
feat: create role assignment and eligibility schedule request resources

josh-barker:
Hi, this is the first time I'm using the new resource generation to create the following resources:

- azurerm_role_assignment_schedule_request
- azurerm_role_eligibility_schedule_request

I'm also happy to create the resources manually in the provider if these resource have some generation issues.

### DIFF
--- a/config/resources/authorizationservice.hcl
+++ b/config/resources/authorizationservice.hcl
@@ -1,0 +1,21 @@
+service "Authorization" {
+  terraform_package = "authorization"
+
+  api "2020-10-01" {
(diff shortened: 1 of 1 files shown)


josh-barker:
Hey @tombuildsstuff , I've been doing some investigation and I don't think this is a good idea, as these are effective "transaction" requests, which don't align to the CRUD terraform resource definition.

However, I'm starting to manually create some resources around PIM for Azure Resources, and have been talking with @manicminer .

PR #2011, opened 2023-01-19
tools/importer-rest-api-specs: rewriting the Resource ID segments to be consistent

tombuildsstuff:
This PR dynamically renames the `UserSpecifiedSegment` keys to match the preceding Static Segment value where possible - for example a Resource ID in the form `/static/{value}` will be updated to `/static/{staticName}`.

Whilst this may seem pedantic, this fixes the repeated inconsistencies seen here where the API Definitions define the same Resource ID using different segments:

https://github.com/hashicorp/pandora/pull/2010/files#diff-aad1409e4b94d6c75f639fd03d2f0f44ed8672f62ca8944813b05e50f4f48817L15-R15

---

This is going to introduce a lot of changes to the generated SDK.. however this one-time breaking change should mostly remove the Swagger issues we're facing when the Segment Values don't align - which feels like a valuable tradeoff - however some usages may need fixing as a result.

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_inconsistent_swagger_segments.go
+++ b/tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_inconsistent_swagger_segments.go
@@ -1,0 +1,55 @@
+package dataworkarounds
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


jackofallops (approved):
LGTM 👍

PR #2012, opened 2023-01-19
config: ignoring some new api versions for now

tombuildsstuff:
We don't need these at this time, so I'm going to ignore these for now


mbfrahry (approved):
LGTM

PR #2030, opened 2023-01-24
tools/importer-rest-api-specs: support for filtering operations outside of the specified Resource Provider

tombuildsstuff:
Some Azure Services define operations outside of their Resource Provider, for example the `network` service includes operations from both `Microsoft.Compute` and `Microsoft.Network` - which causes issues when parsing since duplicate Resource ID names are parsed out (#1223).

This PR adds support for optionally filtering the Resource IDs imported to just those which match the Resource Provider listed in `./config/resource-manager.hcl` - which avoids the issue by filtering these Resource ID's out. Whilst this isn't something we want to do for every Service, doing so allows us to skip these 4 operations which aren't relevant to us - which allows us to import `network`.

Fixes #1223

PR #2033, opened 2023-01-25
tools/importer-rest-api-specs: filtering operations in addition to Resource IDs

tombuildsstuff:
This PR fixes the regression introduced in #2030 which surfaced in #2032, where Resource IDs are removed at a far broader scale than intended.

Since the `ResourceProvider` value is used against the Service MetaData - this PR introduces a new internal-only field for filtering on the Resource Provider - `ResourceProviderToFilterTo` - which ultimately allows for `Network` to be imported/generated.


jackofallops (approved):
LGTM 👍

PR #2091, opened 2023-02-08
Importer: add workaround for batch account autoStorageBaseProperties

stephybun:


tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_batch_21291.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	model, ok := resource.Models["AutoStorageBaseProperties"]
+	if !ok {
+		return nil, fmt.Errorf("couldn't find Model AutoStorageBaseProperties")
+	}
+	field, ok := model.Fields["StorageAccountId"]
+	if !ok {
+		return nil, fmt.Errorf("couldn't find the field StorageAccountId within model AutoStorageProperties")
+	}
+	field.Required = false
```
tombuildsstuff, line 39: we'll also need to set

```suggestion
	field.Optional = true
	field.Required = false
```

tombuildsstuff (commented):
One comment but otherwise 👍

PR #2100, opened 2023-02-09
tools/importer-rest-api-specs: accounting for `azurerm_load_test` being a special-case when generating the examples

tombuildsstuff:
I looked into using `hclwrite` and `hclparse` here, whilst we should take that approach it's significantly more involved and likely wants to instead be done at the same time as the test generation itself - as such for now this is a special-case for the `azurerm_load_test` resource.

Fixes https://github.com/hashicorp/terraform-provider-azurerm/pull/20389


stephybun (approved):
Thanks good job 🥇

PR #2112, opened 2023-02-13
`generator-go-sdk`: ServicesUsingNewBaseLayer support with specific api version

wuxu92:
And add Automation@2022-08-08 with the new base layer. this version has not been used in `azurerm` for the current main and it will fix some SDK issues when working with https://github.com/hashicorp/go-azure-sdk/pull/298.


tombuildsstuff (approved):
LGTM 👍

PR #2122, opened 2023-02-14
tools/generator-go-sdk: update responder code for list methods with discriminated types

stephybun:


tools/generator-go-sdk/generator/templater_methods_autorest.go
```diff
@@ -653,6 +652,66 @@ func (c %[1]s) responderFor%[2]s(resp *http.Response) (result %[6]s, err error)
 		return &output, nil
 	}
 
+	if discriminatedType != "" && strings.HasPrefix(c.operationName, "List") {
```
tombuildsstuff, line 655: not every list operation starts with `List` unfortunately, we'd be better to do:

```suggestion
	if discriminatedType != "" && c.operation.FieldContainingPaginationDetails != nil {
```

tools/generator-go-sdk/generator/templater_methods_autorest.go
```diff
@@ -653,6 +652,66 @@ func (c %[1]s) responderFor%[2]s(resp *http.Response) (result %[6]s, err error)
...
+	var respObj page
+	err = autorest.Respond(
+		resp,
+		%[3]s)
+	result.HttpResponse = resp
+	temp := make([]%[7]s, 0)
+	for _, v := range respObj.Values {
+		val, err := unmarshal%[8]sImplementation(v)
+			if err != nil {
+				err = autorest.NewErrorWithError(err, "%[5]s.%[1]s", "%[2]s", nil, "Failure to unmarshal")
```
tombuildsstuff, line 679: it'd be good to give details for /which/ iteration we've had an unmarshalling error with:

```suggestion
	for i, v := range respObj.Values {
		val, err := unmarshal%[8]sImplementation(v)
			if err != nil {
				err = fmt.Errorf("unmarshalling item %%d for %[8]s (%%q): %+v", i, v, err)
```

tombuildsstuff (commented):
Left a couple of comments inline but this otherwise LGTM 👍

PR #2130, opened 2023-02-16
tools/generator-go-sdk: checking both the service and api version for the new base layer flag

tombuildsstuff:
This PR fixes a bug introduced in #2112 where the Meta Client wasn't being output using the new base layer - since this is output separately to each Resource within an API Version and needs to be output separately.


manicminer (approved):
LGTM 👍

PR #2166, opened 2023-02-24
Config: Add container-instance 2021-09-01

magodo:
Related issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/20604

I'm gonna add a check and use the corresponding client for usgovernment only for the container group resource, but I hesitate about whether that is the right (maintainable) thing to do...

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -84,7 +84,7 @@ service "consumption" {
 }
 service "containerinstance" {
   name      = "ContainerInstance"
-  available = ["2021-10-01", "2022-09-01"]
+  available = ["2021-09-01", "2021-10-01", "2022-09-01"]
 }
 service "containerservice" {
   name      = "ContainerService"


tombuildsstuff:
hey @magodo 

Thanks for this PR.

Taking a look through here, we intentionally avoid this approach of using different API Versions in different Azure Environments since there can be differences in these APIs (both in the payload and also the API behaviour) - meaning the schema and the resource may need to be different.

In this instance whilst this API version (`2021-10-01`) isn't available, slightly older versions are (`2021-03-01`, `2021-07-01` and `2021-09-01`) - and as such it stands to reason that this API version will become available in time in this Azure Environment. As such it'd probably be worth reaching out to the Azure Government team to find out when they plan to deploy this API version - in the interim we'd recommend users pin to the older API Version here when using an Azure Environment other than Azure Public.

When we update the API version being used in the Azure Provider we intentionally call out the API Version upgrades in the changelog ([for example `postgres` here](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/CHANGELOG.md#3440-february-16-2023)) - so it should be possible to determine which version of the Provider should work in this Azure Environment.

Since we're intentionally taking a different approach here, whilst I'd like to thank you for this contribution I'm going to close this PR for the moment.

Thanks!
Tom

PR #2172, opened 2023-02-27
tools/generator-go-sdk: adding the missing copyright lines to the ID Parsers, Tests, Predicates and Meta Clients

tombuildsstuff:
This PR fixes a few issues spotted in https://github.com/hashicorp/go-azure-sdk/pull/337 where the ID Parsers, ID Parser Tests, Predicates and Meta Client files weren't being output with the expected copyright headers.

Once this PR is merged, `hashicorp/go-azure-sdk` will be regenerated to include these changes, meaning that https://github.com/hashicorp/go-azure-sdk/pull/337 will also get regenerated to only target the `.github` folder (which is fine) - so we can close https://github.com/hashicorp/go-azure-sdk/pull/337 in favour of this PR (which, once merged will generate the same fixes)

### DIFF
--- a/tools/generator-go-sdk/generator/templater_id_parser.go
+++ b/tools/generator-go-sdk/generator/templater_id_parser.go
@@ -19,6 +19,11 @@ type resourceIdTemplater struct {
 }
 
 func (r resourceIdTemplater) template(data ServiceGeneratorData) (*string, error) {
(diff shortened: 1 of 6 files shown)


katbyte (approved):
LGTM 🌵

PR #2177, opened 2023-03-01
tools/generator-go-sdk: inverting the list of services using the old …

tombuildsstuff:
…base layer

This means that new services will use the new base layer by default, and we can gradually move off of the older ones


jackofallops (approved):
LGTM 👍

PR #2182, opened 2023-03-02
tools/importer-rest-api-specs: supporting using Common IDs for Key Vault and Key Vault Key

tombuildsstuff:
Should fix the flip-flopping seen in https://github.com/hashicorp/pandora/pull/2181/


stephybun (approved):
LGTM 🥖

PR #2191, opened 2023-03-06
Pull out nested model if it have multiple AllOf

### LINKED ISSUE #1377: `CosmosDB.GremlinGraph` - the struct type of the `Resource` property shouldn't be it's parent type `GremlinGraphGetProperties` but `GremlinGraphResource`
While updating CosmosDB.GremlinGraph resource to use [Pandora sdk](https://github.com/hashicorp/go-azure-sdk/tree/main/resource-manager/cosmosdb/2022-05-15/cosmosdb), I found the struct type of the "Resource" property in the response returned by "[client.GremlinResourcesGetGremlinGraph](https://github.com/hashicorp/go-azure-sdk/blob/6ebf10b92cf136af85307534d0d69e425aeaff80/resource-manager/cosmosdb/2022-05-15/cosmosdb/method_gremlinresourcesgetgremlingraph_autorest.go#L20)" is "[GremlinGraphGetProperties](https://github.com/hashicorp/go-azure-sdk/blob/6ebf10b92cf136af85307534d0d69e425aeaff80/resource-manager/cosmosdb/2022-05-15/cosmosdb/model_gremlingraphgetproperties.go#L6)" so that it is always referring it's parent struct type "[GremlinGraphGetProperties](https://github.com/hashicorp/go-azure-sdk/blob/6ebf10b92cf136af85307534d0d69e425aeaff80/resource-manager/cosmosdb/2022-05-15/cosmosdb/model_gremlingraphgetproperties.go#L6)". So we cannot get the sub properties of "Resource" in read operation. If my understanding is correct, I assume the struct type of the "Resource" property should be "[GremlinGraphResource](https://github.com/Azure/azure-rest-api-specs/blob/ee364f685feedb2ad905c682b57117f9006f541f/specification/cosmos-db/resource-manager/Microsoft.DocumentDB/stable/2022-05-15/cosmos-db.json#L5980)". Could you take a look this issue? Thanks in advance.

Note: Seems the [ThroughputSettingsGetProperties.Resource](https://github.com/hashicorp/go-azure-sdk/blob/fad68f98f363b8cdb1177a8fdb18682047ee2d51/resource-manager/cosmosdb/2022-05-15/cosmosdb/model_throughputsettingsgetproperties.go#L7) and the [GremlinDatabaseGetProperties.Resource](https://github.com/hashicorp/go-azure-sdk/blob/fad68f98f363b8cdb1177a8fdb18682047ee2d51/resource-manager/cosmosdb/2022-05-15/cosmosdb/model_gremlindatabasegetproperties.go#L8) and [SqlContainerGetProperties.Resource](https://github.com/hashicorp/go-azure-sdk/blob/fad68f98f363b8cdb1177a8fdb18682047ee2d51/resource-manager/cosmosdb/2022-05-15/cosmosdb/model_sqlcontainergetproperties.go) have also same problem.

image

wuxu92:
resolves #1377

I have tried to generate the SDK for CosmosDB and it looks fine.

image

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -156,7 +156,8 @@ func (d *SwaggerDefinition) detailsForField(modelName string, propertyName strin
 		result.Append(*nestedResult)
 	}
 
-	if len(value.Properties) > 0 || len(value.AllOf) > 0 {
+	// if there are more than 1 allOf, it can not use a simple reference type, but a new definition
+	if len(value.Properties) > 0 || len(value.AllOf) > 1 {
 		// there's a nested model we need to pull out
 		inlinedName := inlinedModelName(modelName, propertyName)
 		nestedFields := make(map[string]models.FieldDetails, 0)
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -156,7 +156,7 @@ func (d *SwaggerDefinition) detailsForField(modelName string, propertyName strin
 		result.Append(*nestedResult)
 	}
 
-	if len(value.Properties) > 0 {
+	if len(value.Properties) > 0 || len(value.AllOf) > 0 {
 		// there's a nested model we need to pull out
 		inlinedName := inlinedModelName(modelName, propertyName)
 		nestedFields := make(map[string]models.FieldDetails, 0)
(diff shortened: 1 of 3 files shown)


katbyte (commented):
Would it be possible to add a test for this?

stephybun (dismissed):
Thanks for this @wuxu92, this looks fine but could we add a test to validate this change?

wuxu92:
> Thanks for this @wuxu92, this looks fine but could we add a test to validate this change?

A unit test was added.

tombuildsstuff (approved):
LGTM - thanks for pushing the unit test @wuxu92

PR #2220, opened 2023-03-10
tools/generator-terraform: output the sdk client using the new base layer/add an error to the old base layer method

tombuildsstuff:
This PR adds support for the generated resources using either `Azure/go-autorest` or `hashicorp/go-azure-sdk` by conditionally calling the correct Meta Client.

This'll need a little coordination in `terraform-provider-azurerm`, since [the `buildAutoClients` method](https://github.com/hashicorp/terraform-provider-azurerm/blob/6c991f7808ad32cfd77d085f604d08cad584069e/internal/clients/client.go#L260) will need to raise the error.


mbfrahry (approved):
LGTM!

PR #2279, opened 2023-03-27
config: add monitor 2020-10-01 and 2021-07-01-preview

catriona-m:


stephybun (approved):

teowa:
Hi, maybe `2022-06-01` should not be removed, as is used by [data Collection Rule](https://github.com/hashicorp/terraform-provider-azurerm/blob/0c0ed0346d59746663b4122b5d2e638c6b61c28e/internal/services/monitor/monitor_data_collection_rule_resource.go#L14).

PR #2280, opened 2023-03-27
tools/generator-go-sdk: normalizing the value of String constants during unmarshalling

### LINKED ISSUE #417: generator-go-sdk: unmarshaling should normalise the casing of Enum's

tombuildsstuff:
Skimming #758 I realised there's a far simpler way of doing this - hence this PR.

This PR introduces a custom unmarshal function for string constants, meaning that when we unmarshal the JSON response from the API, we'll try and normalize this value to match the casing defined within the API Specification.

As a hypothetical example, this means that is the API Specification defines `sku` with the values `Standard` and `Premium` - if the API response returns `standard` or `Standard` or `StAnDaRd` then we'll unmarshal this value consistently as `Standard`.

Notably, this change only takes effect at Unmarshal time, since we assume that the payload being provided in the HTTP Response is cased correctly, as in the Terraform Provider, the schema should be validating the value being provided matches the constant values.

Supersedes #758
Fixes #417

### DIFF
--- a/tools/generator-go-sdk/featureflags/flags.go
+++ b/tools/generator-go-sdk/featureflags/flags.go
@@ -5,15 +5,6 @@ import (
 	"strings"
 )
 
-// GenerateNormalizationFunctionsForConstants specifies whether Normalization functions
-// should be generated for Constants. This means that APIs which return Constants in a
-// differing casing to those defined in the SDK will be automatically re-cased during
-// deserialization if they match a defined value. For example `standard` -> `Standard`.
-//
-// Values which are unmatched (e.g. new/undefined values) will be left as-is and just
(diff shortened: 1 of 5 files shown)


stephybun (approved):
LGTM 🥚

PR #2286, opened 2023-03-28
Add `ContainerService` `2023-02-02-preview`

stephybun:


tombuildsstuff (approved):
LGTM 👍 - as discussed offline we should consider removing the older versions too

Issue #2265, opened 2023-03-23
API Management @ 2021-08-01 API-diagnostic ID missing `APIName` segment


xuzhang3:
`API-diagnostic` resource ID, `ApiName`  missed in the ID 

ID in Swagger/Portal: `"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/apis/{apiId}/diagnostics/{diagnosticId}"`(https://github.com/Azure/azure-rest-api-specs/blob/main/specification/apimanagement/resource-manager/Microsoft.ApiManagement/stable/2021-08-01/apimapis.json#L2444)

ID in SDK : `"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/diagnostics/{diagnosticId}"`
 
`

expect:
```go
type ApiDiagnosticId struct {
	SubscriptionId string
	ResourceGroup  string
	ServiceName    string
	ApiName        string
	DiagnosticName string
}
```

Actual：
```go
  type DiagnosticId struct {
	SubscriptionId    string
	ResourceGroupName string
	ServiceName       string
	DiagnosticId      string
}
```

PR #2299, opened 2023-03-30
`importer-rest-api-specs`: have to update the `ParsedOperation.ResourceIdName` when name conflicts in resourceID

wuxu92:
should resolves: #2265

because the importer tool will find the resource id definition by resource id name when process one tag(Line 260):

https://github.com/hashicorp/pandora/blob/64e4990f726856ce548d5ab3c25f941c5f583211/tools/importer-rest-api-specs/components/parser/remove_unused_items.go#L253-L262

the generated SDK as expected:

image

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/resourceids/generate_names.go
+++ b/tools/importer-rest-api-specs/components/parser/resourceids/generate_names.go
@@ -12,11 +12,11 @@ import (
 	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
 )
 
-func (p *Parser) generateNamesForResourceIds(input []models.ParsedResourceId) (*map[string]models.ParsedResourceId, error) {
-	return generateNamesForResourceIds(input, p.logger)
+func (p *Parser) generateNamesForResourceIds(input []models.ParsedResourceId, uri2ResourceId map[string]ParsedOperation) (*map[string]models.ParsedResourceId, error) {
+	return generateNamesForResourceIds(input, p.logger, uri2ResourceId)
 }
 
-func generateNamesForResourceIds(input []models.ParsedResourceId, log hclog.Logger) (*map[string]models.ParsedResourceId, error) {
+func generateNamesForResourceIds(input []models.ParsedResourceId, log hclog.Logger, uri2ResourceId map[string]ParsedOperation) (*map[string]models.ParsedResourceId, error) {
 	// now that we have all of the Resource ID's, we then need to go through and determine Unique ID's for those
 	// we need all of them here to avoid conflicts, e.g. AuthorizationRule which can be a NamespaceAuthorizationRule
 	// or an EventHubAuthorizationRule, but is named AuthorizationRule in both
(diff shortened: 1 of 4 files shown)


tools/importer-rest-api-specs/components/parser/resourceids/generate_names.go
```diff
@@ -12,11 +12,11 @@ import (
 	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
 )
 
-func (p *Parser) generateNamesForResourceIds(input []models.ParsedResourceId) (*map[string]models.ParsedResourceId, error) {
-	return generateNamesForResourceIds(input, p.logger)
+func (p *Parser) generateNamesForResourceIds(input []models.ParsedResourceId, uri2ResourceId map[string]ParsedOperation) (*map[string]models.ParsedResourceId, error) {
```
stephybun, line 15: Can we rename this to 
```suggestion
func (p *Parser) generateNamesForResourceIds(input []models.ParsedResourceId, uriToResourceId map[string]ParsedOperation) (*map[string]models.ParsedResourceId, error) {
```

tools/importer-rest-api-specs/components/parser/resourceids/generate_names.go
```diff
@@ -95,6 +95,8 @@ func generateNamesForResourceIds(input []models.ParsedResourceId, log hclog.Logg
 	}
 
 	// now we need to fix the conflicts
+	// we also have to update the ResourceIdName in OriginalUrisToResourceIDs
+	conflictUniqNames := map[string]struct{}{}
```
stephybun, line 99: And rename this to 
```suggestion
	conflictUniqueNames := map[string]struct{}{}
```

tools/importer-rest-api-specs/components/parser/resourceids/generate_names.go
```diff
@@ -116,6 +119,15 @@ func generateNamesForResourceIds(input []models.ParsedResourceId, log hclog.Logg
 	for k, v := range candidateNamesToUris {
 		key := fmt.Sprintf("%sId", cleanup.NormalizeName(k))
 		outputNamesToUris[key] = v
+
+		if _, ok := conflictUniqNames[k]; ok {
+			for idx, v2 := range uri2ResourceId {
+				if v2.ResourceId != nil && v2.ResourceId.ID() == v.ID() && (v2.ResourceIdName != nil && *v2.ResourceIdName != key) {
```
stephybun, line 125: We should be checking if the resource IDs are matching on a segment per segment basis and there should be a .Matches or .Equal method on the ResourceId that we can use for this.

tools/importer-rest-api-specs/components/parser/resourceids/generate_names_test.go
```diff
@@ -433,7 +433,7 @@ var redisPatchSchedulesResourceId = models.ParsedResourceId{
 }
 
 func TestResourceIDNamingEmpty(t *testing.T) {
-	actualNamesToIds, err := generateNamesForResourceIds([]models.ParsedResourceId{}, hclog.NewNullLogger())
+	actualNamesToIds, err := generateNamesForResourceIds([]models.ParsedResourceId{}, hclog.NewNullLogger(), nil)
```
stephybun, line 436: Can we pass in an empty map in these instead of nil

stephybun (changes requested):
Thanks @wuxu92, aside from the comments left in-line can you please add a unit test to validate this?

tools/importer-rest-api-specs/components/parser/resourceids/generate_names.go
tombuildsstuff: @wuxu92 looks like this one's still pending?

wuxu92: sorry i missed this one. updated jus now.

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @wuxu92

PR #2314, opened 2023-04-03
tools/importer-rest-api-specs:  updating alertrules -> alertRules and logprofiles -> logProfiles

catriona-m:


stephybun (approved):
Thanks @catriona-m LGTM 🍄

PR #2316, opened 2023-04-03
config: Add Network 2022-09-01

JenGoldstrich:
👋 This PR adds the 2022-09-01 Network API

https://learn.microsoft.com/en-us/rest/api/virtual-network/

https://github.com/Azure/azure-rest-api-specs/tree/main/specification/network/resource-manager/Microsoft.Network/stable/2022-09-01

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -315,6 +315,10 @@ service "netapp" {
   name      = "NetApp"
   available = ["2021-10-01", "2022-01-01", "2022-05-01", "2022-09-01"]
 }
+service "network" {
+  name = "Network"
+  available = ["2022-09-01"]
+}
 service "networkfunction" {
   name      = "NetworkFunction"
   available = ["2022-08-01", "2022-11-01"]


stephybun:
Thanks for this PR @JenGoldstrich. 

Network is a rather large service and will affect the amount of time it takes to regenerate the data API and go SDK. We have some performance fixes coming up soon which would mitigate this slow down. As a result we'd like to hold off on merging this for the moment. Without any context on the type of work you're doing could I suggest migrating another service first?

tombuildsstuff:
The foundation for the performance fixes has been merged in #2318 fwiw, later this week we'll look to switch the Data API over to using this newly populated data, which should speed the Data API up considerably and allow us to get this one in without a performance regression.

JenGoldstrich:
Hey @stephybun just for context This is actually the last missing set of services I think that I am missing, for some context I am looking to migrating the https://github.com/hashicorp/packer-plugin-azure/ to this SDK, so I am filling in some missing endpoints, I do not need this urgently, and this can definitely wait without blocking me at all!

The packer plugin uses the following network clients which is why i want to import this to the SDK

```
    SubnetsClient
        List endpoints
    InterfacesClient
        Get, Delete endpoints
    VirtualNetworksClient
        ListAllComplete, Delete endpoints 
    SecurityGroups
        Delete endpoint 
    PublicIPAddresses
        Get, Delete endpoints
```

TL;Dr no rush!

tombuildsstuff:
First part of the perf. changes is in #2389, the follow-up PR should put us in a position where we can get this merged

tombuildsstuff (approved):

PR #2320, opened 2023-04-04
tools/generator-go-sdk: avoid setting optional discriminated types to Required 

catriona-m:


tombuildsstuff (approved):
This should be fine, however I think we'll want to merge this/add this to the Provider after this weeks release?

PR #2330, opened 2023-04-05
tools/generator-go-sdk: conditionally outputting the `parse[Constant]` functions

tombuildsstuff:
These functions are needed in two cases:

1. When the Constant is a String, the `parse[Constant]` function is used during Normalization
2. When the Constant is used within a Resource ID, the `parse[Constant]` function is used within the Resource ID parse function

Whilst we could always output these parse functions, conditionally outputting these reduces the TLOC we're outputting, which keeps the Go Module size of `hashicorp/go-azure-sdk` down.

Fixes https://github.com/hashicorp/go-azure-sdk/actions/runs/4615982722?pr=400

### DIFF
--- a/tools/generator-go-sdk/generator/templater_constant_test.go
+++ b/tools/generator-go-sdk/generator/templater_constant_test.go
@@ -162,9 +162,9 @@ func TestTemplateStringConstant(t *testing.T) {
 		CaseInsensitive: false,
 		Type:            resourcemanager.StringConstant,
 		Values: map[string]string{
-			"Berlin": "berlin",
-			"Oslo":   "oslo",
-			"Sydney": "sydney",
+			"Berlin":   "berlin",
+			"Canberra": "canberra",
+			"Oslo":     "oslo",
 		},
 	}, false, false)
 	if err != nil {
@@ -174,23 +174,23 @@ func TestTemplateStringConstant(t *testing.T) {
 
 const (
 	CapitalBerlin Capital = "berlin"
+	CapitalCanberra Capital = "canberra"
 	CapitalOslo Capital = "oslo"
-	CapitalSydney Capital = "sydney"
 )
 
 func PossibleValuesForCapital() []string {
 	return []string{
 		string(CapitalBerlin),
+        string(CapitalCanberra),
         string(CapitalOslo),
-        string(CapitalSydney),
 	}
 }
 
 func parseCapital(input string) (*Capital, error) {
 	vals := map[string]Capital{
 		"berlin": CapitalBerlin,
+        "canberra": CapitalCanberra,
         "oslo": CapitalOslo,
-        "sydney": CapitalSydney,
 	}
 	if v, ok := vals[strings.ToLower(input)]; ok {
     	return &v, nil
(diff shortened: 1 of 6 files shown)


tombuildsstuff:
The resulting output looks fine, when moved into a new Go file for testing:

```go
package main

import (
	"encoding/json"
	"fmt"
	"log"
	"strings"
)

type Model struct {
	Field Kind `json:"kind"`
}

func main() {
	vals := []struct{
		input string
		expected string
	}{
		{
			input:    `{"kind": "Gen1"}`,
			expected: "Gen1",
		},
		{
			input:    `{"kind": "gen1"}`,
			expected: "Gen1",
		},
		{
			input:    `{"kind": "gen3"}`,
			expected: "gen3",
		},
		{
			input:    `{"kind": null}`,
			expected: "",
		},
	}
	for _, v := range vals {
		var val Model
		if err := json.Unmarshal([]byte(v.input), &val); err != nil {
			log.Fatalf("unmarshal error: %+v", err)
		}
		log.Printf("Got %q / expected %q", string(val.Field), v.expected)
	}
}

type Kind string

const (
	KindGenOne Kind = "Gen1"
	KindGenTwo Kind = "Gen2"
)

func PossibleValuesForKind() []string {
	return []string{
		string(KindGenOne),
		string(KindGenTwo),
	}
}

func (s *Kind) UnmarshalJSON(bytes []byte) error {
	var decoded string
	if err := json.Unmarshal(bytes, &decoded); err != nil {
		return fmt.Errorf("unmarshaling: %+v", err)
	}
	out, err := parseKind(decoded)
	if err != nil {
		return fmt.Errorf("parsing %q: %+v", decoded, err)
	}
	*s = *out
	return nil
}

func parseKind(input string) (*Kind, error) {
	vals := map[string]Kind{
		"gen1": KindGenOne,
		"gen2": KindGenTwo,
	}
	if v, ok := vals[strings.ToLower(input)]; ok {
		return &v, nil
	}

	// otherwise presume it's an undefined value and best-effort it
	out := Kind(input)
	return &out, nil
}
```

outputs:

```
2023/04/05 12:06:07 Got "Gen1" / expected "Gen1"
2023/04/05 12:06:07 Got "Gen1" / expected "Gen1"
2023/04/05 12:06:07 Got "gen3" / expected "gen3"
2023/04/05 12:06:07 Got "" / expected ""
```

tools/generator-go-sdk/generator/templater_constant_test.go
```diff
@@ -76,7 +166,55 @@ func TestTemplateStringConstant(t *testing.T) {
...
+	CapitalBerlin Capital = "berlin"
+	CapitalOslo Capital = "oslo"
+	CapitalSydney Capital = "sydney"
+)
+
+func PossibleValuesForCapital() []string {
+	return []string{
+		string(CapitalBerlin),
+        string(CapitalOslo),
+        string(CapitalSydney),
```
stephybun, line 185: Capital of Australia is Sydney mate

stephybun (approved):
Other than your disappointing lack of knowledge in geography LGTM 🦘

Issue #2185, opened 2023-03-02
automation: `ReplaceContent` should send and receive a `RawFile`


tombuildsstuff:
Per the Swagger:

https://github.com/Azure/azure-rest-api-specs/blob/0b1805b3ce076549ba4e9285a4f0f8727caf53a5/specification/automation/resource-manager/Microsoft.Automation/stable/2022-08-08/runbook.json#L130-L136

Today we parse out:

```csharp
namespace Pandora.Definitions.ResourceManager.Automation.v2022_08_08.RunbookDraft;

internal class ReplaceContentOperation : Operations.PutOperation
{
    public override string? ContentType() => "text/powershell";

    public override IEnumerable<HttpStatusCode> ExpectedStatusCodes() => new List<HttpStatusCode>
        {
                HttpStatusCode.Accepted,
                HttpStatusCode.OK,
        };

    public override bool LongRunning() => true;

    public override Type? RequestObject() => typeof(object);

    public override ResourceID? ResourceId() => new RunbookId();

    public override Type? ResponseObject() => typeof(CustomTypes.RawFile);

    public override string? UriSuffix() => "/draft/content";


}
```

However we should be both sending and receiving a `RawFile` type, to give us a consistent Marshal and Unmarshal function - so this is both a feature request for `hashicorp/go-azure-sdk` (to support these Marshal/Unmarshal funcs for `RawFile`'s`) and a bug in the Importer where we don't detect this is a `RawFile` payload (instead we're outputting it as an `object` payload), which whilst semantically the same means we can't rely on this being a `RawFile`, which'll have other issues down the road.

PR #2332, opened 2023-04-06
`importer`:  use `RawFile` instead of `object` for request as object type but format with file

wuxu92:
try to resolve: #2185 

imported cs file:

image

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -638,7 +638,7 @@ func (d SwaggerDefinition) parseNativeType(input *spec.Schema) *models.ObjectDef
 		}
 	}
 
-	if input.Type.Contains("file") {
+	if input.Type.Contains("file") || (input.Format == "file" && input.Type.Contains("object")) {
 		return &models.ObjectDefinition{
 			Type: models.ObjectDefinitionRawFile,
 		}
@@ -656,9 +656,6 @@ func (d SwaggerDefinition) parseNativeType(input *spec.Schema) *models.ObjectDef
 		}
 	}
 
-	// TODO: Objects with Format of File are actually RawFiles..
-	// "type": "object",
-	// "format": "file"
 	if input.Type.Contains("object") {
 		return &models.ObjectDefinition{
 			Type: models.ObjectDefinitionRawObject,
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -638,7 +638,7 @@ func (d SwaggerDefinition) parseNativeType(input *spec.Schema) *models.ObjectDef
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -638,7 +638,7 @@ func (d SwaggerDefinition) parseNativeType(input *spec.Schema) *models.ObjectDef
 		}
 	}
 
-	if input.Type.Contains("file") {
+	if input.Type.Contains("file") || (input.Format == "file" && input.Type.Contains("object")) {
```
tombuildsstuff, line 641: we need this to be case-insensitive due to inconsistencies in the Swagger

```suggestion
	if input.Type.Contains("file") || (strings.EqualsFold(input.Format, "file") && input.Type.Contains("object")) {
```

tools/importer-rest-api-specs/components/parser/testdata/model_single_with_rawfile.json
```diff
@@ -0,0 +1,125 @@
...
+              "format": "file"
+            },
+            "description": "Wrapper class."
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Success.",
+            "schema": {
+              "$ref": "#/definitions/Example"
```
tombuildsstuff, line 53: we should update the Response to return a file too, so that we can verify (in the test above) that we parse this out in both directions

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -656,9 +656,6 @@ func (d SwaggerDefinition) parseNativeType(input *spec.Schema) *models.ObjectDef
 		}
 	}
 
-	// TODO: Objects with Format of File are actually RawFiles..
-	// "type": "object",
-	// "format": "file"
 	if input.Type.Contains("object") {
 		return &models.ObjectDefinition{
```
tombuildsstuff, line 660: thinking about it, it'd make more sense to move this logic here, leaving the change above as-is and making this:

```suggestion
		if strings.EqualsFold(input.Format, "file") {
			return &models.ObjectDefinition{
				Type: models.ObjectDefinitionRawFile,
			}
		}
		return &models.ObjectDefinition{
```

tombuildsstuff (changes requested):
hey @wuxu92 

Thanks for this PR - I've taken a look through and left some comments inline, if we can fix those up then this should otherwise be good to go 👍 

Thanks!

tombuildsstuff (changes requested):
hey @wuxu92 

Thanks for pushing those changes, I've taken a look through and left another couple of minor comments inline, but it appears that some of the comments from last time were missed?

Thanks!

wuxu92:
> 

thanks for your comments, sorry i did not get all your points in the last push 🔆

tombuildsstuff (approved):
LGTM - thanks for pushing those changes @wuxu92

PR #2388, opened 2023-04-17
data: removing definitions which have been removed from the Swagger

tombuildsstuff:
These have been imported from the Swagger but are no longer present within it (due to bugfixes etc), as such these should be removed.

@stephybun is working on the medium-term fix to remove this data at import time too.


stephybun (approved):
LGTM 👍

PR #2389, opened 2023-04-17
Data API: Constants and Models are now loaded from the Resource Definition rather than Dynamically

tombuildsstuff:
This PR dramatically improves the performance of the Data API, such that we go from the `/v1/resource-manager/services/MachineLearningServices/2022-10-01` endpoint taking `25278.9166ms` -> `172.9626ms`.

The first commit updates the Data API to load the Constants and Models from the Resource Definition, which increases the performance from:

```
before:
Request finished HTTP/1.1 GET http://localhost:5000/v1/resource-manager/services/MachineLearningServices/2022-10-01 - - - 200 - application/json;+charset=utf-8 25278.9166ms

after:
Request finished HTTP/1.1 GET http://localhost:5000/v1/resource-manager/services/MachineLearningServices/2022-10-01 - - - 200 - application/json;+charset=utf-8 16318.8919ms
```

The second commit then extends this by removing the reflection logic used when parsing models, such that we can bring this down to `172.9626ms`.

Whilst I'd originally planned to split this out, in truth these may as well be grouped together - meaning what was going to become multiple PR's should be consolidated into this one.


manicminer (approved):
LGTM

PR #2398, opened 2023-04-18
importer-rest-api-specs: remove the api version directories for regeneration and remove empty ones 

stephybun:


tools/importer-rest-api-specs/components/dataapigenerator/helpers.go
```diff
@@ -40,20 +42,27 @@ func fileExistsAtPath(filePath string) (*bool, error) {
 	return &result, nil
 }
 
-func recreateDirectoryExcludingFiles(directory string, excludeList []string, logger hclog.Logger) error {
+func recreateDirectoryExcludingFiles(directory string, logger hclog.Logger) error {
```
tombuildsstuff, line 45: this is no longer excluding files, so the method name should change?

tools/importer-rest-api-specs/components/dataapigenerator/helpers.go
```diff
@@ -40,20 +42,27 @@ func fileExistsAtPath(filePath string) (*bool, error) {
...
 	logger.Trace(fmt.Sprintf("(Re)creating directory at path %q..", directory))
-	if err := os.MkdirAll(directory, os.FileMode(0755)); err != nil {
-		return fmt.Errorf("creating directory %q: %+v", directory, err)
+
+	// if the directory is an api version delete the entire directory and recreate it along with the parent service for regeneration
+	if match, err := regexp.MatchString(`/v20\d\d_\d\d_\d\d(_preview)?`, directory); match && err == nil {
```
tombuildsstuff, line 49: if we move the Service generation to before the API Version generation, we should be able to remove this and use the other `recreateDirectory` method instead?

tombuildsstuff: also worth noting, API versions unfortunately don't _have_ to be date-based (see data plane), even if this is specific to Resource Manager currently

tombuildsstuff (changes requested):
Whilst this approach works, we can probably simplify this by outputting the Service before recreating the API Version?

stephybun:
@tombuildsstuff changed approach and now recreate the entire service directory before we begin parsing the swagger and generating the data

tools/importer-rest-api-specs/components/discovery/interface.go
```diff
@@ -2,6 +2,8 @@ package discovery
 
 import "github.com/hashicorp/pandora/tools/sdk/config/definitions"
 
+const RootNamespace = "Pandora.Definitions.ResourceManager"
```
tombuildsstuff, line 5: this intentionally isn't a constant btw, since it's intended to only be used in this one place since this is specific to Resource Manager - when we add support for the MS Graph definitions, that'll likely become `Pandora.Definitions.MicrosoftGraph`

tombuildsstuff (approved):
One comment but otherwise 👍

PR #2410, opened 2023-04-20
tools/importer-rest-api-specs: passing the Resource IDs for this Reso…

tombuildsstuff:
…urce into the "check for usages of constant" function

Else all constants that are used in a Resource ID are always output, which is unneeded.


jackofallops (approved):
LGTM 👍

PR #2418, opened 2023-04-24
Avoid replacing the discriminated variant's model by its parent's model

magodo:
Currently, there are a couple of Swagger model definitions that have fields whose type is a discriminated variant.

The api importer currently replace those fields' types by using their parents'. The motivation (per my understanding) is to use the parent's `Unmarshaller` during JSON decoding.

The replacement here is actually not necessary, even it breaks the API contract a bit. In fact, we can just avoid doing this replacement in the importer. Additionally, we shall update the SDK generator to only generate the `unmarshalXXXImplementation` for the discriminated parent type. For the variant type, just use its `Unmarshal` method if existed, otherwise, use the default one is enough.

## Impact Analysis

Counting the changed files of the *data*:

```shell
pandora on  avoid_replace_discriminated_variant_by_parent_model
💤  diff -qr -x obj -x bin data-old data | wc -l
92
```

Counting the changed files of the *generated-sdk-dev*:

```shell
💢  diff -rq generated-sdk-dev-old generated-sdk-dev | wc -l
88
```

The files are almost 1:1 mapping, except 4 files in the data that are corresponding to a pure abstract class, e.g. `AutoMLVerticalModel` at *pandora/data/Pandora.Definitions.ResourceManager/MachineLearningServices/v2023_04_01/Schedule/Model-AutoMLVertical.cs*.

An example of the change in the *generated-sdk-dev*:

```shell
diff --color -r generated-sdk-dev-old/resource-manager/datafactory/2018-06-01/credentials/model_managedidentitycredentialresource.go generated-sdk-dev/resource-manager/datafactory/2018-06-01/credentials/model_managedidentitycredentialresource.go
3,7d2
< import (
<       "encoding/json"
<       "fmt"
< )
<
12,45c7,11
<       Etag       *string    `json:"etag,omitempty"`
<       Id         *string    `json:"id,omitempty"`
<       Name       *string    `json:"name,omitempty"`
<       Properties Credential `json:"properties"`
<       Type       *string    `json:"type,omitempty"`
< }
<
< var _ json.Unmarshaler = &ManagedIdentityCredentialResource{}
<
< func (s *ManagedIdentityCredentialResource) UnmarshalJSON(bytes []byte) error {
<       type alias ManagedIdentityCredentialResource
<       var decoded alias
<       if err := json.Unmarshal(bytes, &decoded); err != nil {
<               return fmt.Errorf("unmarshaling into ManagedIdentityCredentialResource: %+v", err)
<       }
<
<       s.Etag = decoded.Etag
<       s.Id = decoded.Id
<       s.Name = decoded.Name
<       s.Type = decoded.Type
<
<       var temp map[string]json.RawMessage
<       if err := json.Unmarshal(bytes, &temp); err != nil {
<               return fmt.Errorf("unmarshaling ManagedIdentityCredentialResource into map[string]json.RawMessage: %+v", err)
<       }
<
<       if v, ok := temp["properties"]; ok {
<               impl, err := unmarshalCredentialImplementation(v)
<               if err != nil {
<                       return fmt.Errorf("unmarshaling field 'Properties' for 'ManagedIdentityCredentialResource': %+v", err)
<               }
<               s.Properties = &impl
<       }
<       return nil
---
>       Etag       *string                   `json:"etag,omitempty"`
>       Id         *string                   `json:"id,omitempty"`
>       Name       *string                   `json:"name,omitempty"`
>       Properties ManagedIdentityCredential `json:"properties"`
>       Type       *string                   `json:"type,omitempty"`
```

This matches its definition in Swagger: https://github.com/Azure/azure-rest-api-specs/blob/2b079c913809722148b25d5fa6577142478cdba2/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/datafactory.json#L7303-L7306.

## E2E test

With above change, replacing the go module to using the updated code in the provider. Additionally, we need to modify the code a bit to make it matches the types, e.g.:

```
diff --git a/internal/services/monitor/monitor_metric_alert_resource.go b/internal/services/monitor/monitor_metric_alert_resource.go
index 22bff1e875..2f6f99f4bd 100644
--- a/internal/services/monitor/monitor_metric_alert_resource.go
+++ b/internal/services/monitor/monitor_metric_alert_resource.go
@@ -601,7 +601,7 @@ func expandMonitorMetricAlertCriteria(d *pluginsdk.ResourceData, isLegacy bool)
 }

 func expandMonitorMetricAlertSingleResourceMultiMetricCriteria(input []interface{}) metricalerts.MetricAlertCriteria {
-       criteria := make([]metricalerts.MultiMetricCriteria, 0)
+       criteria := make([]metricalerts.MetricCriteria, 0)
        for i, item := range input {
                v := item.(map[string]interface{})
                dimensions := expandMonitorMetricDimension(v["dimension"].([]interface{}))
@@ -736,11 +736,11 @@ func flattenMonitorMetricAlertCriteria(input metricalerts.MetricAlertCriteria) [
        }
 }

-func flattenMonitorMetricAlertSingleResourceMultiMetricCriteria(input *[]metricalerts.MultiMetricCriteria) []interface{} {
+func flattenMonitorMetricAlertSingleResourceMultiMetricCriteria(input *[]metricalerts.MetricCriteria) []interface{} {
        if input == nil || len(*input) == 0 {
                return nil
        }
-       criteria := (*input)[0].(metricalerts.MetricCriteria)
+       criteria := (*input)[0]
        metricName := criteria.MetricName
        metricNamespace := criteria.MetricNamespace
        timeAggregation := criteria.TimeAggregation
```

(Note this is actually a revert to the changes done during the transition from Azure SDK to `go-azure-sdk`)

Then run the tests that cover the impacted properties:

```shell
馃挗  TF_ACC=1 go test -v -timeout=20h ./internal/services/monitor -run=TestAccMonitorMetricAlert_basic
=== RUN   TestAccMonitorMetricAlert_basic
=== PAUSE TestAccMonitorMetricAlert_basic
=== RUN   TestAccMonitorMetricAlert_basicAndCompleteUpdate
=== PAUSE TestAccMonitorMetricAlert_basicAndCompleteUpdate
=== CONT  TestAccMonitorMetricAlert_basic
=== CONT  TestAccMonitorMetricAlert_basicAndCompleteUpdate
--- PASS: TestAccMonitorMetricAlert_basic (253.32s)
--- PASS: TestAccMonitorMetricAlert_basicAndCompleteUpdate (754.35s)
PASS
ok      github.com/hashicorp/terraform-provider-azurerm/internal/services/monitor       754.363s
```


tombuildsstuff:
@magodo apologies for the delayed review - FWIW this conceptually looks fine, this is just waiting for a good time to merge this given there's breaking changes involved as a result of this (and there's already some pending in the Provider in https://github.com/hashicorp/terraform-provider-azurerm/pull/21910) - but this otherwise should be fine.

So that we can get this merged, would you mind removing the changes to the `./data/Pandora.Definitions.ResourceManager/` folder? Those get regenerated after a PR merge, so typically conflict as new things get generated (as above)

magodo:
@tombuildsstuff Sure, I've overwrite the `data` folder from main branch and merge the main branch as well.

tombuildsstuff:
@magodo can you revert the `data` changes entirely? Given this merge needs to be coordinated it's likely this'll conflict again otherwise - having the `./data/Pandora.Definition.ResourceManager/` path changed only through auto-generation solves this fwiw

PR #2445, opened 2023-04-27
tools/generator-go-sdk: disabling / feature-flagging the discriminator change from #2320 for now

tombuildsstuff:
This change is impactful and needs to be rolled out Service Package by Service Package, as such we're going to temporarily disable this to allow us to update the SDK, and then once that's done we can look to enable this Service by Service by using the `new base layer` feature toggle to let us make those changes.


stephybun (approved):
LGTM 🚲

PR #2509, opened 2023-05-15
tools/generator-go-sdk: porting more services over to using the new base layer

tombuildsstuff:
Also removed a bunch of services which are generated but aren't currently used in `terraform-provider-azurerm` - there's no need to handle the base layer migration for them since these can be migrated cleanly


jackofallops (commented):
LGTM 👍

PR #2462, opened 2023-05-02
config: ignore `network` `2022-11-01`

stephybun:
Ignoring until a few SDK clients from network have been tested

PR #2510, opened 2023-05-16
Config: add `network@2022-11-01`

ziyeqf:
This fix only patched `2022-11-01` https://github.com/Azure/azure-rest-api-specs/pull/23738

The ignore was added on https://github.com/hashicorp/pandora/pull/2462, not sure if it has been unblocked.

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -337,8 +337,7 @@ service "netapp" {
 }
 service "network" {
   name      = "Network"
-  available = ["2022-09-01"]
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
As [per this comment](https://github.com/tombuildsstuff/kermit/pull/192#discussion_r1195835704) this is ignored because it was unused, I don't believe there's anything blocking this - so we'll merge both the Kermit and go-azure-sdk version of this for now.

tombuildsstuff (approved):

PR #2540, opened 2023-05-23
tools/generator-terraform: outputting `identity` as blocks in the attributes/arguments - and updating the `block reference`

### LINKED ISSUE #2270: for azurerm_load_test identity - update documentation - input a block not a string & extend to allow UserAssigned
documentation is currently missing some readability details - input for `identity` should be a block containing type.
> https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/load_test

currently:
* `identity` - (Optional) Specifies the Managed Identity which should be assigned to this Load Test.

should be:
```
* `identity` - (Optional) An `identity` block as defined below.
---
An `identity` block supports the following:
* `type` - (Required) The type of Managed Identity used for this Load Test. Possible values are `SystemAssigned`.
```

also
requesting the identity aligns with AzureRM template and accepts both types of Managed Identities

> `type` | Type of managed service identity (where both SystemAssigned and UserAssigned types are allowed).
> `identity_ids` | The set of user assigned identities associated with the resource. etc
> https://learn.microsoft.com/en-us/azure/templates/microsoft.loadtestservice/loadtests?pivots=deployment-language-terraform

(as an example)
There is a similar identity block available for AzureRM Automation Account - allowing 3 options:

> An identity block supports the following:
> `type` - (Required) The type of identity used for this Automation Account. Possible values are SystemAssigned, UserAssigned and SystemAssigned, UserAssigned.
> `identity_ids`- (Optional) The ID of the User Assigned Identity which should be assigned to this Automation Account.
> Note:
> identity_ids is required when type is set to UserAssigned or SystemAssigned, UserAssigned.
> https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/automation_account

support case 104926 
also feature request 1204255964099438 

thanks - Sharon (HashiCorp Terraform Support)

tombuildsstuff:
This PR does a few things:

* [x] Outputs the `identity` fields as Blocks for both Arguments and Attributes
* [x] Outputs the `identity` blocks inline within the Blocks Reference
* [x] Outputs nested blocks with context of where these are located within the Blocks Reference when there's conflicts (e.g. `the "bar" block within the "foo" block supports:`) to account for blocks having the same name at different locations (also de-duping when these are the same)
* [x] Outputting Lists and Sets appropriately ("One or more XXX blocks" and "Each XXX block supports..")
* [x] Outputting the location of each block (e.g. "the XXX block as defined (above|below)")

Fixes #2270

### DIFF
--- a/tools/generator-terraform/generator/resource/docs/component_arguments.go
+++ b/tools/generator-terraform/generator/resource/docs/component_arguments.go
@@ -72,7 +72,6 @@ func getArguments(model resourcemanager.TerraformSchemaModelDefinition, resource
 	return &out, nil
 }
 
-// TODO: perhaps we should have specific tests covering each line (to validate things like above/below/validation etc?)
 func documentationLineForArgument(field resourcemanager.TerraformSchemaFieldDefinition, nestedWithin, resourceName string) (*string, error) {
 	components := make([]string, 0)
 	components = append(components, fmt.Sprintf("* `%s` -", field.HclName))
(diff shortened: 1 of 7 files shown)


jackofallops (approved):

PR #2545, opened 2023-05-24
add MsSql 2022-11-01-preview

katbyte:


mbfrahry (approved):
LGTM!

config/resource-manager.hcl
```diff
@@ -468,6 +468,10 @@ service "solutions" {
   name      = "ManagedApplications"
   available = ["2019-07-01", "2021-07-01"]
 }
+service "sql" {
+  name      = "MsSql"
```
tombuildsstuff, line 472: can we keep this as `Sql` rather than `MsSql` here, because of `SqlVirtualMachine` below?

```suggestion
  name      = "Sql"
```

PR #2560, opened 2023-05-26
tools/importer-rest-api-specs: fixing an eventual consistency bug when parsing the data

tombuildsstuff:
This occurs when the map of keys hasn't been populated yet, thus depending on the ordering this may or may not get flattened, causing issues


manicminer (approved):
LGTM 👍

PR #2567, opened 2023-05-30
update `batch` storageAccountId workaround for `2022-10-01`

myc2h6o:
to unblock https://github.com/hashicorp/terraform-provider-azurerm/pull/21962


stephybun (approved):
Thanks @myc2h6o LGTM 👍

PR #2576, opened 2023-05-31
` tools/terraform-generator`: field transform compatible with legacy identity

wuxu92:
current transformer functions cannot support legacy identity models. add logic to support direct assignment of legacy identity block in terraform generator.

### DIFF
--- a/tools/generator-terraform/generator/mappings/assignment_direct.go
+++ b/tools/generator-terraform/generator/mappings/assignment_direct.go
@@ -54,8 +54,15 @@ func (d directAssignmentLine) assignmentForCreateUpdateMapping(mapping resourcem
 	}
 
 	// check if it requires a custom transform
-	if v, ok := transformTypes[schemaField.ObjectDefinition.Type]; ok && v == sdkField.ObjectDefinition.Type {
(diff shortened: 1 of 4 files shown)


tombuildsstuff:
hey @wuxu92 

Thanks for this PR - apologies for the delayed review here.

I've been digging into what's needed to update Load Test to the latest stable version of the API and whilst the changes proposed in this PR go some way towards that, unfortunately the design direction here needs to be rather different.

Namely, rather than shoe-horning support for the Legacy Identity types in - since it's possible for one Schema Type to map to multiple times in the SDK we'll need to change this backing data type from a Map to something else, which has been done in https://github.com/hashicorp/pandora/pull/2590.

There's also some other changes needed too, namely the `encryption` block is going to require threading a few things through - in order to facilitate that we're going to temporarily remove this block whilst that work's going on (which is being tracked in https://github.com/hashicorp/pandora/issues/2608) and then once that's fixed up we'll look to reintroduce this field.

As such whilst I'd like to thank you for this PR, since this work has been superseded by https://github.com/hashicorp/pandora/pull/2590 I'm going to close this PR for the moment - but the work to update to the latest API version has been rolled out in https://github.com/hashicorp/pandora/pull/2609.

Thanks!

PR #2583, opened 2023-06-02
`tools/importer-rest-api-specs`: refactoring Terraform Acceptance Test generation

tombuildsstuff:
This PR refactors the Rest API Specs Importer such that the Terraform Acceptance Test generation now accounts for both Attributes and Blocks (and recurses these as needed). It also adds support for dynamically determining more dependent resources - notably including the Key Vault and Key Vault Key resources, which enables fixing #2549.

At this point in time this switches to outputting Terraform Variables rather than using the `locals` block - this is something we'll change in the future, but in the short-term this'll make dynamically outputting the Terraform Variables in the Terraform Generator easier.

There's more to do for this work to be usable - namely the Terraform Generator also needs updating to account for this, but since I've reworked the `testattributes` package into more granular `testing` package, this seemed like a good stopping point (which, once reviewed will need to wait for the Terraform Generator changes prior to merging).


jackofallops (dismissed):

stephybun (approved):
Blobby exception

PR #2590, opened 2023-06-05
tools/generator-terraform: direct assignments now support transformations of `SystemAndUserAssignedIdentity` (Schema) -> `LegacySystemAndUserAssignedIdentity` (SDK)

tombuildsstuff:
This PR refactors the `terraform-generator` tool to support one Schema Type (in this case `SystemAndUserAssignedIdentity`) being mapped to multiple SDK Types (in this case `SystemAndUserAssignedIdentityMap` and `LegacySystemAndUserAssignedIdentityMap`) - opting to split this into structs since this is no longer suitable as a map.

Since we're looking to upgrade the Load Test Service to a Stable API Version - this PR includes a temporary hack to remove a field which is problematic (namely the `encryption` block) since some additional changes are needed within the `importer-rest-api-specs`, `Data API`, `generator-go-sdk` and `generator-terraform` to support Data Plane Key Vault Key IDs - which will be tracked in a separate issue.

Supersedes #2576
Supersedes #2497


tombuildsstuff:
TODO: open a summary issue for Load Test

tools/importer-rest-api-specs/components/schema/processors/field_name_rename_mislabelled_resource_id_test.go
```diff
@@ -0,0 +1,180 @@
+package processors
+
+import (
+	"github.com/hashicorp/go-azure-helpers/lang/pointer"
+	"testing"
+
+	"github.com/hashicorp/pandora/tools/sdk/resourcemanager"
```
jackofallops, line 7: goimports

jackofallops (approved):
Thanks @tombuildsstuff - just one goimports comment, otherwise LGTM 👍

PR #2618, opened 2023-06-09
config: remove `PaloAltoNetworks` until resource ID segments are fixed

stephybun:
See #2617

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -374,6 +374,7 @@ service "orbital" {
 service "paloaltonetworks" {
   name      = "PaloAltoNetworks"
   available = ["2022-08-29"]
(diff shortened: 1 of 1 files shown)


config/resource-manager.hcl
```diff
@@ -374,6 +374,7 @@ service "orbital" {
 service "paloaltonetworks" {
   name      = "PaloAltoNetworks"
   available = ["2022-08-29"]
+  ignore    = ["2022-08-29"]
 }
```
tombuildsstuff, line 378: I don't _think_ this works, we'd either need to remove the block for now - or fix the casing and move forward?

stephybun: Removed.

tombuildsstuff (approved):

PR #2639, opened 2023-06-13
version-bumper: trim `-legacy` suffix from api version names

stephybun:


tools/version-bumper/config.go
```diff
@@ -188,6 +188,7 @@ func normalizeDateString(input string) string {
 	output = strings.TrimSuffix(output, "-publicpreview")
 	output = strings.TrimSuffix(output, "-preview")
 	output = strings.TrimSuffix(output, "-beta")
+	output = strings.TrimSuffix(output, "-legacy")
```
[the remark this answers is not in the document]
stephybun: is legacy a preview version though? or are we supposed to treat it like one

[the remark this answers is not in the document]
tombuildsstuff: Followed up offline, but `legacy` is a preview version in this context (this function is used to determine whether a new API version is a stable API version that should get added to the config - so perhaps this should be more accurately repurposed as `IsStableVersion`?

tools/version-bumper/config.go
```diff
@@ -121,7 +121,7 @@ func reconcileWithAvailableServices(input services.Config, availableServices []A
 			}
 
 			// for now we don't want to auto-import Preview versions, only Stable releases
-			if isPreviewVersion(version) {
+			if isStableVersion(version) {
```
tombuildsstuff, line 124:
```suggestion
			if !isStableVersion(version) {
```

tools/version-bumper/config.go
```diff
@@ -160,7 +160,7 @@ func isVersionIsOlderThanExistingVersion(new string, existing string) (*bool, er
 		v := true
 		// if we're using an older preview (e.g. 2020-01-01-preview) but
 		// a GA version gets published, we should import the new version
-		if isPreviewVersion(existing) {
+		if isStableVersion(existing) {
```
tombuildsstuff, line 163:
```suggestion
		if !isStableVersion(existing) {
```

tools/version-bumper/config.go
```diff
@@ -188,13 +188,14 @@ func normalizeDateString(input string) string {
...
 	output = strings.TrimSuffix(output, "-beta")
+	output = strings.TrimSuffix(output, "-legacy")
 	return output
 }
 
 func filterToStableVersions(input []string) []string {
 	output := make([]string, 0)
 	for _, v := range input {
-		if isPreviewVersion(v) {
+		if isStableVersion(v) {
```
tombuildsstuff, line 198:
```suggestion
		if !isStableVersion(v) {
```

tools/version-bumper/config.go
```diff
@@ -205,10 +206,11 @@ func filterToStableVersions(input []string) []string {
 	return output
 }
 
-func isPreviewVersion(version string) bool {
+func isStableVersion(version string) bool {
```
tombuildsstuff, line 209: if we're renaming this function (so it's more accurate) we'll want to flip the true/false below, since right now this is returning _that_ it's a Preview API?

tombuildsstuff (changes requested):
Left a few comments inline, the function will want updating too?

tombuildsstuff (approved):

PR #2642, opened 2023-06-14
config: add graph `2023-04-13`

teowa:
swagger: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/graphservicesprod/resource-manager/Microsoft.GraphServices/stable/2023-04-13/graphservicesprod.json

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -223,6 +223,10 @@ service "frontdoor" {
   name      = "FrontDoor"
   available = ["2020-04-01", "2020-05-01", "2022-05-01"]
 }
+service "graphservicesprod" {
+  name = "GraphServices"
+  available = ["2023-04-13"]
+}
 service "guestconfiguration" {
   name      = "GuestConfiguration"
   available = ["2020-06-25", "2022-01-25"]


tombuildsstuff:
@teowa is this Microsoft Graph? What's this provisioning? cc @manicminer

teowa:
Now only the Microsoft.GraphServices/accounts resource can be provisioned. From [doc](https://learn.microsoft.com/en-us/graph/metered-api-setup?tabs=azurecloudshell#enable-an-application):
> To enable an application to use metered APIs and services in Microsoft Graph, it must be associated with an Azure subscription. To create this association, you must create an Azure resource of type Microsoft.GraphServices/accounts. The Azure resource connects a single Azure AD application registration with the Azure subscription where the application's usage of metered APIs is billed.

manicminer:
@teowa Thanks for suggesting to add this. In general we'd prefer to keep that functionality in the AzureAD TF provider (via MS Graph). Is there functionality available in this RP that is not available by directly querying Microsoft Graph?

manicminer (approved):

config/resource-manager.hcl
```diff
@@ -223,6 +223,10 @@ service "frontdoor" {
   name      = "FrontDoor"
   available = ["2020-04-01", "2020-05-01", "2022-05-01"]
 }
+service "graphservicesprod" {
```
manicminer, line 226:
```suggestion
service "graphservices" {
```

PR #2664, opened 2023-06-15
generator-terraform: add support for parsing the parent resource ID

stephybun:

PR #2679, opened 2023-06-21
importer-rest-api-specs: define common id for kubernetes cluster which is a parent resource

stephybun:
Prerequisite for #2664


tombuildsstuff (approved):
👍 bear in mind https://github.com/hashicorp/go-azure-helpers/pull/171 will need to be merged first then vendored into azurerm & go-azure-sdk for this to be usable end-to-end - but 👍

PR #2734, opened 2023-07-07
data: supporting `ContainerApps` @ `2022-11-01-preview`

tombuildsstuff:
Fixes https://github.com/hashicorp/go-azure-sdk/issues/469 Fixes https://github.com/hashicorp/go-azure-sdk/issues/470

PR #2741, opened 2023-07-10
tools/importer-rest-api-specs: ModelDefinition comparisons are now more detailed and Reference checks are now case-insensitive

tombuildsstuff:
Fixes an issue when parsing ContainerApps@2022-11-01-preview and Web@2022-09-01

Unblocks https://github.com/hashicorp/pandora/pull/2734

I've tried adding an acctest to cover this scenario, but I'm struggling to do so due to the normalization - so I've omitted this for now.


stephybun (approved):
Thanks TW LGTM 🥧

PR #2742, opened 2023-07-10
tools/version-bumper: skipping updating services which are EOL/Deprecated/appear to be deprecated when importing the full set

tombuildsstuff:
These also don't import correctly currently due to Swagger issues, however since they're EOL/Deprecated we can skip them


stephybun (approved):
LGTM 📚

PR #2309, opened 2023-03-31
config: add EventGrid `2022-06-15`

catriona-m:

PR #2763, opened 2023-07-13
`tools/importer-rest-api-specs`: detecting generic ARM Resource scopes

### LINKED ISSUE #502: importer: detect hard-coded scopes as scopes
The swagger contains a series of scopes defined as:

> /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/{providerNamespace}/{resourceTypeName}/{resourceName}

which are essentially Scopes - rather than outputting this ID we should investigate flagging these as Scopes and use that instead?

### LINKED ISSUE #2341: EventGrid: `2021-12-01` sdk generation failing on resourceIds
Generating EventGrid 2021-12-01 gives the following error:

```
Error: -31T09:49:08.366Z [ERROR]      💥 Error: parsing Swagger files: parsing files in "../../../swagger/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2021-12-01": parsing Resource Ids from "/EventGrid.json" (Service "EventGrid" / Api Version "2021-12-01"): finding Resource IDs: mapping Processed Resource IDs to the Input Resource IDs: couldn't find the processed ID Name for Resource URI "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventGrid/{parentType}/{parentName}/privateEndpointConnections"
    run_validate_can_parse_data_test.go:75: error: parsing Swagger files: parsing files in "../../../swagger/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2021-12-01": parsing Resource Ids from "/EventGrid.json" (Service "EventGrid" / Api Version "2021-12-01"): finding Resource IDs: mapping Processed Resource IDs to the Input Resource IDs: couldn't find the processed ID Name for Resource URI "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventGrid/{parentType}/{parentName}/privateEndpointConnections"
```

The parentType here can be either `topics` or `domains`: https://github.com/Azure/azure-rest-api-specs/blob/3a8a03746104623f69c3dccab27ad1ac842a7130/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2021-12-01/EventGrid.json#L2637

tombuildsstuff:
This fixes an issue when importing EventGrid, but detects Generic ARM Resource Scopes which are defined as ARM Resources with User Specified Segments (e.g. `/subscriptions/{}/resourceGroups/{}/providers/{}/{}/{}`) instead of Scopes (`/{scope}`). This is also the case for "nested" items (for example `/subscriptions/{}/resourceGroups/{}/providers/{}/{}/{}/{}/{}`)

These should be normalised into the Common ID `ScopeId` - but aren't captured with the Common IDs detection logic
since these can be one of many slightly different formats, so this is easiest to special-case.

Fixes #2341
Fixes #502
Unblocks #2309

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/resource_ids_test.go
+++ b/tools/importer-rest-api-specs/components/parser/resource_ids_test.go
@@ -297,75 +297,86 @@ func TestParseResourceIdContainingAScope(t *testing.T) {
 }
 
 func TestParseResourceIdContainingAHiddenScope(t *testing.T) {
-	result, err := ParseSwaggerFileForTesting(t, "resource_ids_containing_hidden_scope.json")
-	if err != nil {
-		t.Fatalf("parsing: %+v", err)
+	files := []string{
+		"resource_ids_containing_hidden_scope.json",
+		"resource_ids_containing_hidden_scope_constant.json",
+		"resource_ids_containing_hidden_scope_hardcoded_provider.json",
 	}
-	if result == nil {
-		t.Fatal("result was nil")
-	}
-	if len(result.Resources) != 1 {
-		t.Fatalf("expected 1 resource but got %d", len(result.Resources))
-	}
+	for _, file := range files {
+		t.Run(file, func(t *testing.T) {
+			fileName := file
 
-	hello, ok := result.Resources["Example"]
-	if !ok {
-		t.Fatalf("no resources were output with the tag Example")
-	}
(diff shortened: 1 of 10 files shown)


tombuildsstuff:
NOTE: this'll cause a large number of changes in the SDK, so wants to be rolled out to account for that

PR #2767, opened 2023-07-14
importer-rest-api-specs: ignore fields in existing resources from update payload temporarily

stephybun:
The changes made for generating the new resource trusted access role binding will correctly mark the field `hub_profile` and `description` in the other generated resources as `ForceNew: false`, but the functionality for update tests isn't built out enough to handle generating the update tests for these yet so this temporarily overrides the value for ForceNew until we build out the functionality in the tests.


tombuildsstuff (approved):
🏉

PR #2769, opened 2023-07-14
tools/importer-rest-api-specs: fixing an issue where Resource IDs which are hidden ARM Scopes are pulled out correctly

tombuildsstuff:
This fixes an issue where `/{scope}/{resourceName}` was being output instead of `/{scope}`


stephybun (approved):
LGTM 🧦

Issue #2753, opened 2023-07-12
`Automation`: runbook draft request content-type not match the swagger


wuxu92:
swagger defined as `text/powershell`:

https://github.com/Azure/azure-rest-api-specs/blob/1c23efa3fffb2bba8dab4875e6e6639c596b76e3/specification/automation/resource-manager/Microsoft.Automation/stable/2022-08-08/runbook.json#L103-L110

but the generated sdk request is set to `application/json`:

https://github.com/hashicorp/go-azure-sdk/blob/2fcb9571adac281a917919918a258b3fb4d0bcdd/resource-manager/automation/2022-08-08/runbookdraft/method_replacecontent.go#L24-L28

this makes the body of a `[]byte` be `json.Marshal`-ed as a base64 encoded string. which causes issues.

PR #2799, opened 2023-07-24
tols/generator-go-sdk: outputting the content type if needed

tombuildsstuff:
Fixes https://github.com/hashicorp/pandora/issues/2753

### DIFF
--- a/tools/generator-go-sdk/generator/templater_methods.go
+++ b/tools/generator-go-sdk/generator/templater_methods.go
@@ -496,17 +496,21 @@ func (c methodsPandoraTemplater) requestOptions() (*string, error) {
 		options = "OptionsObject: options,"
 	}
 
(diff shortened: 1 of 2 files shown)


tombuildsstuff:
Example output:

```golang
// ReplaceContent ...
func (c RunbookDraftClient) ReplaceContent(ctx context.Context, id RunbookId, input []byte) (result ReplaceContentOperationResponse, err error) {
	opts := client.RequestOptions{
		ContentType: "text/powershell",
		ExpectedStatusCodes: []int{
			http.StatusAccepted,
			http.StatusOK,
		},
		HttpMethod: http.MethodPut,
		Path:       fmt.Sprintf("%s/draft/content", id.ID()),
	}

	req, err := c.Client.NewRequest(ctx, opts)
	if err != nil {
		return
	}

	if err = req.Marshal(input); err != nil {
		return
	}
```

PR #2837, opened 2023-07-26
Remove 2020-01-01 from pandora

mbfrahry:
We need [legacy 2021-05-01](https://github.com/Azure/azure-rest-api-specs/tree/7c50841aadb4bfc241de3f09146e1f97ce3583b2/specification/mysql/resource-manager/Microsoft.DBforMySQL/legacy/stable/2021-05-01) but we already support a version `2021-05-01`. Is there a way we can point to the legacy version as well?

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -344,7 +344,7 @@ service "msi" {
 }
 service "mysql" {
   name      = "MySql"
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
@mbfrahry `legacy` `2021-05-01` is just `2021-05-01`, so we can just use that instead 👍

tombuildsstuff (approved):

PR #2866, opened 2023-08-01
Importer: Custom filed user assigned identity (map) tolerate exposed principal id at top level

magodo:
The storage cache AML file system resource defined UAI with the `principalId` and `tenantId` defined at the top identity model level: https://github.com/Azure/azure-rest-api-specs/blob/14d24d17491d8c2bde24532cb8cc2d663c0ffd9f/specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2023-05-01/amlfilesystem.json#L565.

Without this change, the identity field of the `AMLFileSystem` model will not be defined as the commonschema.

Related to: https://github.com/hashicorp/terraform-provider-azurerm/pull/22680

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/commonschema/custom_field_identity_user_assigned_map.go
+++ b/tools/importer-rest-api-specs/components/parser/commonschema/custom_field_identity_user_assigned_map.go
@@ -95,9 +95,9 @@ func (userAssignedIdentityMapMatcher) IsMatch(_ models.FieldDetails, definition 
 			continue
 		}
 
-		// some services expose tenant ID in the definition for a user assigned identity - we should recognise it as a user assigned identity but ignore the field
-		// https://github.com/Azure/azure-rest-api-specs/blob/5a9afce8360020c46b38841e04179447a28118b2/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2022-12-01/FlexibleServers.json#L812-L834
-		if strings.EqualFold(fieldName, "TenantId") {
+		// some services expose tenant ID or/and principal ID in the definition for a user assigned identity - we should recognise it as a user assigned identity but ignore the field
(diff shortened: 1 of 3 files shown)


tools/importer-rest-api-specs/components/parser/commonschema/custom_field_identity_user_assigned_map.go
```diff
@@ -95,9 +95,9 @@ func (userAssignedIdentityMapMatcher) IsMatch(_ models.FieldDetails, definition
 			continue
 		}
 
-		// some services expose tenant ID in the definition for a user assigned identity - we should recognise it as a user assigned identity but ignore the field
-		// https://github.com/Azure/azure-rest-api-specs/blob/5a9afce8360020c46b38841e04179447a28118b2/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2022-12-01/FlexibleServers.json#L812-L834
-		if strings.EqualFold(fieldName, "TenantId") {
+		// some services expose tenant ID or/and principal ID in the definition for a user assigned identity - we should recognise it as a user assigned identity but ignore the field
+		// https://github.com/Azure/azure-rest-api-specs/blob/14d24d17491d8c2bde24532cb8cc2d663c0ffd9f/specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2023-05-01/amlfilesystem.json#L565
+		if strings.EqualFold(fieldName, "TenantId") || strings.EqualFold(fieldName, "PrincipalId") {
```
tombuildsstuff, line 100: Could we link to both for posterity / add a test covering this too?

tombuildsstuff (changes requested):
hey @magodo 

Thanks for this PR - taking a look through it'd be good to add a test to cover this scenario too, since we need to ensure we're handling a rogue `TenantId` field being present as well as both the `PrincipalId` and `TenantId` fields being present - but if we can fix that up this otherwise LGTM 👍

magodo:
@tombuildsstuff Thank you for the instant review! I've updated the comment and add a new test case to cover the `principalId`, please take another look!

tombuildsstuff (approved):
LGTM - thanks for this @magodo

PR #2869, opened 2023-08-01
tools/generator-go-sdk: exporting the Raw generated types

tombuildsstuff:
This PR fixes https://github.com/hashicorp/go-azure-sdk/issues/287 by exposing the Raw types rather than using internal types for these.

Since these fields are intentionally only consumable via the Response and are not usable in Request Payloads exposing this is reasonable, but I've added a warning that these should be a last resort/workaround only.


manicminer (approved):
LGTM 👍

PR #2887, opened 2023-08-03
refactor: moving the `swagger` submodule to `submodules/rest-api-specs`

tombuildsstuff:
This matches the Microsoft Graph one, and since we're gonna need to run `git submodule init` we might as well do this now too


jackofallops (approved):
LGTM. 👍

PR #2891, opened 2023-08-03
tooling: fixing an issue where the branch name was duplicated

tombuildsstuff:
This means the first job to complete would win, the other would fail - and we'd not import all the data


jackofallops (approved):
LGTM 👍

PR #2960, opened 2023-08-21
config: Add container registry version 2023-06-01-preview

magodo:
Adding container registry version 2023-06-01-preview. This is to update the ACR API from 2021-08-01-preview, that is used by the provider.


tombuildsstuff:
@magodo #2976 just added a new Stable API version `2023-07-01`, can the Stable API Version be used instead?

magodo:
@tombuildsstuff There is a property for ACR named `anonymousPullEnabled` that is only available in the preview API. So that's why I targeted to that version instead of the stable one.

PR #3005, opened 2023-09-01
config: adding API Versions for Application Insights

tombuildsstuff:
These two versions are currently used in the provider


manicminer (approved):
LGTM

PR #3017, opened 2023-09-06
tools/importer-rest-api-specs: fixing a bug where operation names may start with a rogue `s`

tombuildsstuff:
Fixes https://github.com/hashicorp/pandora/pull/3016#pullrequestreview-1612987765

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/parser.go
+++ b/tools/importer-rest-api-specs/components/parser/parser.go
@@ -92,7 +92,15 @@ func (d *SwaggerDefinition) simplifyOperationNamesForResource(resource models.Az
 	output := make(map[string]models.OperationDetails)
 	for key, value := range resource.Operations {
 		updatedKey := key[len(resourceNameLower):]
-		// trim off any spurious `s` at the start, to handle tags with singular vs plural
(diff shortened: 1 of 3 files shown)


tools/importer-rest-api-specs/components/parser/parser.go
```diff
@@ -92,6 +92,11 @@ func (d *SwaggerDefinition) simplifyOperationNamesForResource(resource models.Az
 	output := make(map[string]models.OperationDetails)
 	for key, value := range resource.Operations {
 		updatedKey := key[len(resourceNameLower):]
+		// trim off any spurious `s` at the start, to handle tags with singular vs plural
+		if strings.HasPrefix(updatedKey, "s") {
+			updatedKey = updatedKey[1:]
+		}
```
stephybun, line 98: Comment implies tags is the only instance where we might have a spurious 's', in which case I think code should be updated to address only that case (or comment updated/clarified)

stephybun (approved):
I have a question, otherwise LGTM ⚡

tools/importer-rest-api-specs/components/parser/parser.go
tombuildsstuff: tags here == Swagger Tags not the literal value `tags` - but I'll update that to clarify that

PR #3036, opened 2023-09-12
importer-rest-api-specs: add output directory for YAML and output API and Service definitions

stephybun:


tools/importer-rest-api-specs/cmd/segments.go
```diff
@@ -50,7 +50,7 @@ func (c SegmentsCommand) Run(_ []string) int {
 		DataApiEndpoint:          nil,
 		JustOutputSegments:       true,
 		Logger:                   logger,
-		OutputDirectory:          os.DevNull,
+		OutputDirectoryCS:        os.DevNull,
```
jackofallops, line 53: Do we need `/dev/null/` in here for the yaml path for the point we remove the CS?

tools/importer-rest-api-specs/components/dataapigeneratoryaml/definition_service.go
```diff
@@ -0,0 +1,43 @@
...
+
+	code, err := codeForServiceDefinition(s.namespaceForService, s.serviceName, s.resourceProvider, s.terraformPackageName, apiVersions)
+	if err != nil {
+		return fmt.Errorf("marshaling Service Definition: %+v", err)
+	}
+	if err := writeYamlToFile(serviceDefinitionFilePath, code); err != nil {
+		return fmt.Errorf("generating Service Definition into %q: %+v", serviceDefinitionFilePath, err)
+	}
+
+	// ServiceDefinition-GenerationSettings.cs is retained between regenerations, so we special-case it
```
jackofallops, line 22: minor: comment refers to cs files

tools/importer-rest-api-specs/components/dataapigeneratoryaml/revision_id.go
```diff
@@ -0,0 +1,25 @@
+package dataapigeneratoryaml
+
+import (
+	"fmt"
+	"gopkg.in/yaml.v3"
```
jackofallops, line 5: We might want this to be? (It's the same, but via the typical route, rather than gopkg)
```suggestion
	"github.com/go-yaml/yaml"
```
ref: https://pkg.go.dev/gopkg.in/yaml.v3#pkg-overview

jackofallops (approved):
Thanks @stephybun - apart from the couple of nits I accidentally called out in single comments somehow, rather this review, this LGTM 👍

PR #2675, opened 2023-06-21
`DevCenter` @ `2022-04-01`

jiaweitao001:
Onboarding Azure DevCenter. 

Swagger: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/devcenter/resource-manager/Microsoft.DevCenter/stable/2023-04-01

PR #3041, opened 2023-09-14
`tools/importer`: skip build model if input schema is only for property

wuxu92:
should fix the importer issue of #2675. also cherry-picked @tombuildsstuff 's commit of adding Unit Test for it: https://github.com/hashicorp/pandora/commit/d2ea0bbda48287bccd3aa705a7832879ae8b9542

```
--- PASS: TestParseModelBug2675DuplicateModel (0.02s)
PASS
```

I have regenerated all services and it caused no different from main branch's data, (except the DevCenter service):

image

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -148,7 +148,7 @@ func (d *SwaggerDefinition) detailsForField(modelName string, propertyName strin
 	}
 
 	// first get the object definition
-	objectDefinition, nestedResult, err := d.parseObjectDefinition(modelName, propertyName, &value, result)
+	objectDefinition, nestedResult, err := d.parseObjectDefinition(modelName, propertyName, &value, result, false)
 	if err != nil {
 		return nil, nil, fmt.Errorf("parsing object definition: %+v", err)
 	}
@@ -451,7 +451,13 @@ func (d *SwaggerDefinition) findAncestorType(input spec.Schema) (*string, *strin
 	return nil, nil, nil
 }
 
-func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string, input *spec.Schema, known internal.ParseResult) (*models.ObjectDefinition, *internal.ParseResult, error) {
+// if `inputForModel` is false, it means the `input` schema cannot be used to parse the model of `modelName`
+func (d SwaggerDefinition) parseObjectDefinition(
+	modelName, propertyName string,
+	input *spec.Schema,
+	known internal.ParseResult,
+	isInputForModel bool,
+) (*models.ObjectDefinition, *internal.ParseResult, error) {
 	// find the object and any models and constants etc we can find
 	// however _don't_ look for discriminator implementations - since that should be done when we're completely done
 	result := internal.ParseResult{
@@ -509,7 +515,7 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 		}
 
 		// then call ourselves to work out what to do with it
-		objectDefinition, nestedResult, err := d.parseObjectDefinition(*objectName, propertyName, topLevelObject, knownIncludingPlaceholder)
+		objectDefinition, nestedResult, err := d.parseObjectDefinition(*objectName, propertyName, topLevelObject, knownIncludingPlaceholder, true)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -526,11 +532,13 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 		// then just return that object instead, there's no point creating the wrapper type
 		if len(input.Properties) == 0 && len(input.AllOf) == 1 {
 			inheritedModel := input.AllOf[0]
-			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result)
+			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result, true)
 		}
 
 		// check for / avoid circular references
-		if _, ok := result.Models[modelName]; !ok {
+		// only parse model when modelName equals PropertyName, otherwise the `input` Schema may not for the modelName
+		// if parsing a top-level or an inlined model, should pass the propertyName just the same as the modelName
+		if _, ok := result.Models[modelName]; !ok && isInputForModel {
 			nestedResult, err := d.parseModel(modelName, *input)
 			if err != nil {
 				return nil, nil, fmt.Errorf("parsing object from inlined model %q: %+v", modelName, err)
@@ -568,7 +576,7 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 			innerModelName = input.AdditionalProperties.Schema.Title
 		}
 
-		nestedItem, nestedResult, err := d.parseObjectDefinition(innerModelName, propertyName, input.AdditionalProperties.Schema, result)
+		nestedItem, nestedResult, err := d.parseObjectDefinition(innerModelName, propertyName, input.AdditionalProperties.Schema, result, true)
 		if err != nil {
 			return nil, nil, fmt.Errorf("parsing nested item for dictionary: %+v", err)
 		}
@@ -591,7 +599,7 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 			inlinedName = fmt.Sprintf("%s%sInlined", cleanup.NormalizeName(modelName), cleanup.NormalizeName(propertyName))
 		}
 
-		nestedItem, nestedResult, err := d.parseObjectDefinition(inlinedName, propertyName, input.Items.Schema, result)
+		nestedItem, nestedResult, err := d.parseObjectDefinition(inlinedName, propertyName, input.Items.Schema, result, true)
 		if err != nil {
 			return nil, nil, fmt.Errorf("parsing nested item for array: %+v", err)
 		}
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -148,7 +148,8 @@ func (d *SwaggerDefinition) detailsForField(modelName string, propertyName strin
 	}
 
 	// first get the object definition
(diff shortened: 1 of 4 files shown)


tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -451,7 +451,13 @@ func (d *SwaggerDefinition) findAncestorType(input spec.Schema) (*string, *strin
 	return nil, nil, nil
 }
 
-func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string, input *spec.Schema, known internal.ParseResult) (*models.ObjectDefinition, *internal.ParseResult, error) {
+// if `inputForModel` is false, it means the `input` schema cannot be used to parse the model of `modelName`
+func (d SwaggerDefinition) parseObjectDefinition(
+	modelName, propertyName string,
+	input *spec.Schema,
+	known internal.ParseResult,
+	isInputForModel bool,
```
tombuildsstuff, line 459: since this is being used to determine if we're passing a model, or pulling inlined information from a field - this is probably better named as:

```suggestion
	parsingModel bool,
```

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -509,7 +515,7 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 		}
 
 		// then call ourselves to work out what to do with it
-		objectDefinition, nestedResult, err := d.parseObjectDefinition(*objectName, propertyName, topLevelObject, knownIncludingPlaceholder)
+		objectDefinition, nestedResult, err := d.parseObjectDefinition(*objectName, propertyName, topLevelObject, knownIncludingPlaceholder, true)
```
tombuildsstuff, line 518: presumably this should be passing `parsingModel` in when looping around? else we'll potentially be passing `true` when the `false` thing:

```suggestion
		objectDefinition, nestedResult, err := d.parseObjectDefinition(*objectName, propertyName, topLevelObject, knownIncludingPlaceholder, parsingModel)
```

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -526,11 +532,13 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 		// then just return that object instead, there's no point creating the wrapper type
 		if len(input.Properties) == 0 && len(input.AllOf) == 1 {
 			inheritedModel := input.AllOf[0]
-			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result)
+			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result, true)
```
tombuildsstuff, line 535: (as above)

```suggestion
			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result, parsingModel)
```

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -568,7 +576,7 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 			innerModelName = input.AdditionalProperties.Schema.Title
 		}
 
-		nestedItem, nestedResult, err := d.parseObjectDefinition(innerModelName, propertyName, input.AdditionalProperties.Schema, result)
+		nestedItem, nestedResult, err := d.parseObjectDefinition(innerModelName, propertyName, input.AdditionalProperties.Schema, result, true)
```
tombuildsstuff, line 579: (as above)

```suggestion
		nestedItem, nestedResult, err := d.parseObjectDefinition(innerModelName, propertyName, input.AdditionalProperties.Schema, result, parsingModel)
```

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -591,7 +599,7 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
 			inlinedName = fmt.Sprintf("%s%sInlined", cleanup.NormalizeName(modelName), cleanup.NormalizeName(propertyName))
 		}
 
-		nestedItem, nestedResult, err := d.parseObjectDefinition(inlinedName, propertyName, input.Items.Schema, result)
+		nestedItem, nestedResult, err := d.parseObjectDefinition(inlinedName, propertyName, input.Items.Schema, result, true)
```
tombuildsstuff, line 602: (as above)

```suggestion
		nestedItem, nestedResult, err := d.parseObjectDefinition(inlinedName, propertyName, input.Items.Schema, result, parsingModel)
```

tools/importer-rest-api-specs/components/parser/operations.go
```diff
@@ -417,7 +417,7 @@ func (p operationsParser) requestObjectForOperation(input parsedOperation, known
 
 	for _, param := range unexpandedOperation.Parameters {
 		if strings.EqualFold(param.In, "body") {
-			objectDefinition, result, err := p.swaggerDefinition.parseObjectDefinition(param.Schema.Title, param.Schema.Title, param.Schema, known)
+			objectDefinition, result, err := p.swaggerDefinition.parseObjectDefinition(param.Schema.Title, param.Schema.Title, param.Schema, known, true)
```
tombuildsstuff, line 420: to make this clearer, could we pull this out so it's more obvious what this does:

```suggestion
			parsingModel := true
			objectDefinition, result, err := p.swaggerDefinition.parseObjectDefinition(param.Schema.Title, param.Schema.Title, param.Schema, known, parsingModel)
```

tools/importer-rest-api-specs/components/parser/operations.go
```diff
@@ -473,7 +473,7 @@ func (p operationsParser) responseObjectForOperation(input parsedOperation, know
 			continue
 		}
 
-		objectDefinition, nestedResult, err := p.swaggerDefinition.parseObjectDefinition(details.ResponseProps.Schema.Title, details.ResponseProps.Schema.Title, details.ResponseProps.Schema, result)
+		objectDefinition, nestedResult, err := p.swaggerDefinition.parseObjectDefinition(details.ResponseProps.Schema.Title, details.ResponseProps.Schema.Title, details.ResponseProps.Schema, result, true)
```
tombuildsstuff, line 476: (same here)

```suggestion
		parsingModel := true
		objectDefinition, nestedResult, err := p.swaggerDefinition.parseObjectDefinition(details.ResponseProps.Schema.Title, details.ResponseProps.Schema.Title, details.ResponseProps.Schema, result, parsingModel)
```

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -526,11 +532,13 @@ func (d SwaggerDefinition) parseObjectDefinition(modelName, propertyName string,
...
 		if len(input.Properties) == 0 && len(input.AllOf) == 1 {
 			inheritedModel := input.AllOf[0]
-			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result)
+			return d.parseObjectDefinition(inheritedModel.Title, propertyName, &inheritedModel, result, true)
 		}
 
 		// check for / avoid circular references
-		if _, ok := result.Models[modelName]; !ok {
+		// only parse model when modelName equals PropertyName, otherwise the `input` Schema may not for the modelName
+		// if parsing a top-level or an inlined model, should pass the propertyName just the same as the modelName
```
tombuildsstuff, line 540: I think this'd be clearer as:

```suggestion
		// however we should only do this when we're parsing a model (`parsingModel`) directly rather than when parsing a model from a field - and only if we haven't already parsed this model
```

tools/importer-rest-api-specs/components/parser/models.go
```diff
@@ -148,7 +148,7 @@ func (d *SwaggerDefinition) detailsForField(modelName string, propertyName strin
 	}
 
 	// first get the object definition
-	objectDefinition, nestedResult, err := d.parseObjectDefinition(modelName, propertyName, &value, result)
+	objectDefinition, nestedResult, err := d.parseObjectDefinition(modelName, propertyName, &value, result, false)
```
tombuildsstuff, line 151: It's hard to understand exactly what the bool is doing here, so if we assign that to an intermediate variable it's far clearer:

```suggestion
	parsingModel  := false
	objectDefinition, nestedResult, err := d.parseObjectDefinition(modelName, propertyName, &value, result, parsingModel)
```

tools/importer-rest-api-specs/components/parser/models_test.go
```diff
@@ -1861,3 +1861,20 @@ func TestParseModelWithLocation(t *testing.T) {
...
+	if result == nil {
+		t.Fatal("result was nil")
+	}
+
+	example, ok := result.Resources["Example"]
+	if !ok {
+		t.Fatal("the Resource 'Example' was not found")
+	}
+
+	t.Logf("%d", len(example.Operations))
```
tombuildsstuff, line 1879: this commit was intended as a quick POC to prove the problem, would you mind adding a few checks to ensure that we're pulling out the correct number of models/that the models we're expecting actually exist?

tombuildsstuff (changes requested):
hey @wuxu92 

Thanks for this PR - I've taken a look and left some comments inline, but this is looking pretty good - if we can fix those up then this should otherwise be good to go 👍 

Thanks!

tools/importer-rest-api-specs/components/parser/models.go
wuxu92: I think we should always pass a `true` here, because the Schame represented by `topLevelObject` is what the `objectName` want. that means we can use `topLevelObject` to parse model `objectName`. also as other two calls below, the inner/inheritedModel is corresponding to the modelName, and they should always pass a `true`.

tools/importer-rest-api-specs/components/parser/models_test.go
wuxu92: updated.

tombuildsstuff (commented):
hi @wuxu92 

If we can fix up the remaining comment (about the comment) then this should otherwise be good to go 👍 

Thanks!

wuxu92:
@tombuildsstuff Thanks a lot for reminding me of missing one comment ❤️ . updated now.

tombuildsstuff (approved):
LGTM 👍

PR #3042, opened 2023-09-14
importer-rest-api-specs: add templates for generating package definition, constants and resourceids in yaml

stephybun:


tools/importer-rest-api-specs/components/dataapigeneratoryaml/templater_constants.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func codeForConstant(constantName string, details resourcemanager.ConstantDetails) ([]byte, error) {
+	sortedKeys := make([]string, 0)
+	for key := range details.Values {
+		sortedKeys = append(sortedKeys, key)
+	}
+	sort.Strings(sortedKeys)
+
+	var values []Value
```
tombuildsstuff, line 30: in the case there's no constants, `values` would be nil here which has a different value to empty when marshalled:

```suggestion
	values := make([]Value, 0)
```

tools/importer-rest-api-specs/components/dataapigeneratoryaml/templater_constants.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func mapConstantFieldType(input resourcemanager.ConstantType) (*string, error) {
+	result := func(in string) (*string, error) {
+		return &in, nil
+	}
```
tombuildsstuff, line 63: can we use `pointer.To` here?

tools/importer-rest-api-specs/components/dataapigeneratoryaml/templater_package_definition.go
```diff
@@ -0,0 +1,66 @@
...
+	for operation := range input.Operations {
+		operationNames = append(operationNames, operation)
+	}
+	sort.Strings(operationNames)
+
+	operations := make([]string, 0)
+
+	for _, operationName := range operationNames {
+		operations = append(operations, fmt.Sprintf("%sOperation", operationName))
+	}
```
tombuildsstuff, line 29: we don't need to suffix these now, since we're outputting these into different files/subsections and not using reflection to find these - meaning we can just use `operationNames` here?

```suggestion
	operationNames := make([]string, 0)
	for operation := range input.Operations {
		operationNames = append(operationNames, operation)
	}
	sort.Strings(operationNames)
```

tools/importer-rest-api-specs/components/dataapigeneratoryaml/templater_package_definition.go
```diff
@@ -0,0 +1,66 @@
...
+	constantNames := make([]string, 0)
+	for k := range input.Constants {
+		constantNames = append(constantNames, k)
+	}
+	sort.Strings(constantNames)
+
+	constants := make([]string, 0)
+	for _, constantName := range constantNames {
+		constants = append(constants, fmt.Sprintf("%sConstant", constantName))
+	}
```
tombuildsstuff, line 40: (same here) we don't need to suffix these anymore, that was a workaround for reflection/all of these being output into the same c# namespace:

```suggestion
```

tools/importer-rest-api-specs/components/dataapigeneratoryaml/templater_package_definition.go
```diff
@@ -0,0 +1,66 @@
...
+	modelNames := make([]string, 0)
+	for k := range input.Models {
+		modelNames = append(modelNames, k)
+	}
+	sort.Strings(modelNames)
+
+	resourceModels := make([]string, 0)
+	for _, modelName := range modelNames {
+		resourceModels = append(resourceModels, fmt.Sprintf("%sModel", modelName))
+	}
```
tombuildsstuff, line 51: (same here)

```suggestion
```

tools/importer-rest-api-specs/components/dataapigeneratoryaml/templater_resource_id.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+	case resourcemanager.UserSpecifiedSegment:
+		{
+			segment.Type = "UserSpecified"
+			segment.Name = input.Name
+
+			return &segment, nil
+		}
+	}
```
tombuildsstuff, line 128: We don't really gain anything by assigning a variable for `segment` here?

```suggestion
	switch input.Type {
	case resourcemanager.ConstantSegment:
		{
			if input.ConstantReference == nil {
				return nil, fmt.Errorf("constant segment type with missing constant reference: %+v", input)
			}

			return &Segment{
				Type: "Constant",
				Name: input.Name,
				Value: *input.ConstantReference
			}, nil
		}

	case resourcemanager.ResourceGroupSegment:
		{
			return &Segment{
				Type: "ResourceGroup",
				Name: resourceGroupName
			}, nil
		}

	case resourcemanager.ResourceProviderSegment:
		{
			if input.FixedValue == nil {
				return nil, fmt.Errorf("resource provider segment type with missing fixed value: %+v", input)
			}

			return &Segment{
				Type: "ResourceProvider",
				Name: input.Name,
				Value: *input.FixedValue
			}, nil
		}

	case resourcemanager.ScopeSegment:
		{
			return &Segment{
				Type: "Scope",
				Name: input.Name,
			}, nil
		}

	case resourcemanager.StaticSegment:
		{
			if input.FixedValue == nil {
				return nil, fmt.Errorf("static segment type with missing fixed value: %+v", input)
			}

			return &Segment{
				Type: "Static",
				Name: input.Name,
				Value: *input.FixedValue
			}, nil

		}

	case resourcemanager.SubscriptionIdSegment:
		{
			return &Segment{
				Type: "SubscriptionId",
				Name: input.Name
			}, nil
		}

	case resourcemanager.UserSpecifiedSegment:
		{
			return &Segment{
				Type: "UserSpecified",
				Name: input.Name
			}, nil
		}
	}
```

tombuildsstuff (approved):
Left a few comments inline but otherwise 👍

PR #3134, opened 2023-10-10
F/add templates for terraform

mbfrahry:
- [x] [ResourceIDShortening](https://github.com/hashicorp/pandora/compare/f/add-templates-for-terraform?expand=1#diff-1aef066d5b4a48753bfc853280ae96fd647aae31b0c5e2cbe2dfb24f455d6ac9R41) `ToResourceIdSegmentNamed` and `ToCommonIdSegmentNamed` seems long winded and that `ResourceId`/`CommonId` would be better I'm wondering if that naming is tied specifically to C# or are we using it for something else? 
- [x] [ObjectDefinition](https://github.com/hashicorp/pandora/compare/f/add-templates-for-terraform?expand=1#diff-1aef066d5b4a48753bfc853280ae96fd647aae31b0c5e2cbe2dfb24f455d6ac9R41) This hasn't been hooked up yet as I haven't figured out how deep to go if it's a reference to a reference to a reference
- [ ] The Resource Test Configs are being spit out on one line rather than using the new lines that are already embedded (ie
```
 "Basic": "provider 'azurerm' {\n  features {}\n}\n\nresource 'azurerm_load_test' 'test' {\n  location            = azurerm_resource_group.test.location\n  name                = 'acctestlt-${var.random_string}'\n  resource_group_name = azurerm_resource_group.test.name\n}",

```


tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	updateMethod := Method{
+		Generate:         details.UpdateMethod.Generate,
+		Name:             details.UpdateMethod.MethodName,
+		TimeoutInMinutes: details.UpdateMethod.TimeoutInMinutes,
+	}
+
+	resourceDefinition := ResourceDefinition{
+		DisplayName: details.DisplayName,
+		// todo figure this out
```
stephybun, line 54: Is there still something to figure out here?

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
```diff
@@ -0,0 +1,94 @@
+package dataapigeneratorjson
+
+import (
+	"encoding/json"
+	"github.com/hashicorp/pandora/tools/sdk/resourcemanager"
+)
+
+type ResourceMapping struct {
+	ResourceIdMapping             *[]ResourceIdMapping            `json:"ResourceIdMapping"`
```
stephybun, line 9: I think we want to add `omitempty` to the json tag since this is a pointer.

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
```diff
@@ -0,0 +1,94 @@
...
+	DirectAssignmentFieldMappings *[]DirectAssignmentFieldMapping `json:"DirectAssignmentFieldMappings,omitempty"`
+	ModelToModelFieldMappings     *[]ModelToModelFieldMapping     `json:"ModelToModelFieldMappings,omitempty"`
+}
+
+type ResourceIdMapping struct {
+	MappingFunction string
+	SegmentName     string
+}
```
stephybun, line 17: I think this should look more like 

```suggestion
type ResourceIdMapping struct {
	SchemaFieldName    string
	SegmentName        string
	ParsedFromParentID bool
}
```

The `ParsedFromParentID` flag determines whether we use the mapping function `ToCommonIdSegmentNamed` or `ToResourceIdSegmentNamed` in the C# files output by the old `dataapigenerator`. In this case I don't think there's any need to differentiate via a function name, the presence of the flag should give us all the information we need.

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
```diff
@@ -0,0 +1,94 @@
...
+	SdkModelName    string `json:"SdkModelName"`
+	SdkFieldPath    string `json:"SdkFieldPath"`
+}
+
+type ModelToModelFieldMapping struct {
+	SchemaModelName string `json:"SchemaModelName"`
+	SdkModelName    string `json:"SdkModelName"`
+	SdkFieldName    string `json:"SdkFieldName"`
+}
+
```
stephybun, line 31: I think there might be one more struct missing for `ModelToModelMapping`

```suggestion
type ModelToModelMapping struct {
	SchemaModelName string `json:"SchemaModelName"`
	SdkModelName string `json:"SdkModelName"`
}
```

tools/importer-rest-api-specs/featureflags/constants.go
```diff
@@ -6,4 +6,4 @@ package featureflags
 const AllowConstantsWithoutXMSEnum = true
 
 // GenerateJsonDataAPI specifies whether the Data API in JSON should be generated.
-const GenerateJsonDataAPI = false
```
stephybun, line 9: Can we keep this turned off for now - at least until we've done a manual review of it together e.g. generate a couple of services/scenarios and take a look at the output.

stephybun (changes requested):
Thanks for these @mbfrahry! Other than a few comments I think this looks like a good first iteration.

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
mbfrahry: Yes! That's pointing to the wrong thing. Let me take a look

tools/importer-rest-api-specs/featureflags/constants.go
mbfrahry: 💯

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
mbfrahry: Fixed and added DirectAssignmentFieldMapping/ModelToModelFieldMapping as they are slightly different and it's probably worth adding that distinction

stephybun: I'm not sure if there are some commits missing, but I don't see them

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_schema.go
```diff
@@ -0,0 +1,104 @@
...
+
+type SchemaModel struct {
+	Fields []SchemaField `json:"Fields"`
+}
+
+type SchemaField struct {
+	HclName       string                           `json:"HclName"`
+	Type          string                           `json:"Type"`
+	Documentation *string                          `json:"Documentation,omitempty"`
+	Required      *bool                            `json:"Required,omitempty"`
```
stephybun, line 20: ObjectDefinition was removed but I think this is still needed. Perhaps we leave it as a TODO and take a look at it together?

mbfrahry: Got it figured out and added!

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_tests.go
```diff
@@ -0,0 +1,48 @@
...
+	"strings"
+
+	"github.com/hashicorp/pandora/tools/sdk/resourcemanager"
+)
+
+type TerraformResourceTestConfig struct {
+	Basic          string  `json:"Basic"`
+	RequiresImport string  `json:"RequiresImport"`
+	CompleteConfig *string `json:"CompleteConfig,omitempty"`
+	TemplateConfig *string `json:"TemplateConfig,omitempty"`
```
tombuildsstuff, line 14: it's possible for us to have other tests here too, so we need to support those:

```suggestion
	TemplateConfig *string `json:"TemplateConfig,omitempty"`
	OtherTests *map[string]string `json:"OtherTests,omitempty"`
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
tombuildsstuff: Not sure what happened here, but these are above now.

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
```diff
@@ -0,0 +1,84 @@
...
+	readMethod := Method{
+		Generate:         details.ReadMethod.Generate,
+		Name:             details.ReadMethod.MethodName,
+		TimeoutInMinutes: details.ReadMethod.TimeoutInMinutes,
+	}
+	updateMethod := Method{
+		Generate:         details.UpdateMethod.Generate,
+		Name:             details.UpdateMethod.MethodName,
+		TimeoutInMinutes: details.UpdateMethod.TimeoutInMinutes,
+	}
```
tombuildsstuff, line 53: resources aren't guaranteed to have an `UpdateMethod`, so this would crash here

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+type ResourceDefinition struct {
+	DisplayName                  string `json:"DisplayName"`
+	Id                           string `json:"Id"`
+	Label                        string `json:"Label"`
+	Category                     string `json:"Category"`
+	Description                  string `json:"Description"`
+	ExampleUsage                 string `json:"ExampleUsage"`
+	GenerateIDValidationFunction bool   `json:"GenerateIDValidationFunction"`
```
tombuildsstuff, line 18: we're using `Id` throughout the rest of the codebase (for fields - methods use `ID` to differentiate them) - so can we update this to match:

```suggestion
	GenerateIdValidationFunction bool   `json:"GenerateIdValidationFunction"`
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
```diff
@@ -0,0 +1,84 @@
...
+		ReadMethod:                   readMethod,
+		UpdateMethod:                 updateMethod,
+	}
+
+	id, ok := resourceIds[details.ResourceIdName]
+	if !ok {
+		return nil, fmt.Errorf("could not find id %s for Terraform Resource %s", details.ResourceIdName, resourceLabel)
+	}
+
+	resourceDefinition.Id = id.String()
```
tombuildsstuff, line 76: `String()` is the Display Value (e.g. `Virtual Machine (Name %q / Resource Group %q)` - this wants to be `.ID()`

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
```diff
@@ -0,0 +1,91 @@
+package dataapigeneratorjson
+
+import (
+	"encoding/json"
+	"github.com/hashicorp/pandora/tools/sdk/resourcemanager"
```
tombuildsstuff, line 5: goimports:

```suggestion

	"github.com/hashicorp/pandora/tools/sdk/resourcemanager"
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
```diff
@@ -0,0 +1,91 @@
...
+type ResourceMapping struct {
+	ResourceIdMapping             *[]ResourceIdMapping            `json:"ResourceIdMapping,omitempty"`
+	DirectAssignmentFieldMappings *[]DirectAssignmentFieldMapping `json:"DirectAssignmentFieldMappings,omitempty"`
+	ModelToModelFieldMappings     *[]ModelToModelFieldMapping     `json:"ModelToModelFieldMappings,omitempty"`
+}
+
+type ResourceIdMapping struct {
+	SchemaFieldName    string `json:"SchemaFieldName"`
+	SegmentName        string `json:"SegmentName"`
+	ParsedFromParentID bool   `json:"ParsedFromParentID"`
```
tombuildsstuff, line 17: (as above)

```suggestion
	ParsedFromParentId bool   `json:"ParsedFromParentId"`
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_schema.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/hashicorp/pandora/tools/sdk/resourcemanager"
+)
+
+type SchemaModel struct {
+	Fields []SchemaField `json:"Fields"`
+}
+
+type SchemaField struct {
+	HclName          string                           `json:"HclName"`
+	Type             string                           `json:"Type"`
```
tombuildsstuff, line 18: `Type` should no longer be present since we've got `ObjectDefinition`

```suggestion
```

tombuildsstuff (approved):
hey @mbfrahry 

Apologies for the delayed review of this one.

I've left some comments inline, whilst on the whole this is looking ok there's some cleanup needed (e.g. we should document the fields) - and also a few bugs which need to be fixed.
As such I'm going to suggest that we merge this PR and then address these issues (and some of the other issues within this package) in a follow-up PR.

PR #3135, opened 2023-10-10
docs: documenting the different types of Object Definitions

tombuildsstuff:


docs/internals-object-definition.md
```diff
@@ -0,0 +1,123 @@
...
+* Float (the Swagger defines Double/Decimal/Number - which we normalize into a Float)
+* Integers
+* List
+* Strings
+
+Examples:
+
+* A Boolean
+* A CSV of a String
+* A List of a String
```
stephybun, line 26: Could we provide an actual example of these? like is the CSV `apple,pear,strawberry` but a list of a string is `["apple","pear","strawberry"]`?

docs/internals-object-definition.md
```diff
@@ -0,0 +1,123 @@
...
+* A String
+
+Limitations:
+
+* A List cannot contain a List
+
+Notes:
+
+* At present the Object Definition type is reused, but shouldn't be, as we'll want to extend this in the future when we parse Data Plane (which has way more possible values for types).
```
stephybun, line 35: Could we be explicit about what specifically is missing here? e.g.
* add an Options Object Definition type
* add a Terraform Object Definition type

docs/internals-object-definition.md
```diff
@@ -0,0 +1,123 @@
...
+* A Reference (to a Constant)
+* A Reference (to a Model)
+* A List of a List of a Reference (to a Constant)
+* A List of a List of a Reference (to a Model)
+* A Dictionary of a Dictionary to a String
+* A SystemOrUserAssignedIdentityList
+
+Notes:
+
+* In retrospect this should probably be named `SDKObjectDefinition`
```
stephybun, line 88: Is this a rename that we should undertake? It's unclear what this sentence is trying to convey or add.

tombuildsstuff:
In the end we ended up deciding to document these inline within the new SDK types, so I'm going to close this out:

* [SDK Object Definition](https://github.com/hashicorp/pandora/blob/main/tools/data-api-sdk/v1/models/sdk_object_definition.go) (there's an associated `SDK Object Definition Type` too)
* [SDK Operation Option Object Definition](https://github.com/hashicorp/pandora/blob/main/tools/data-api-sdk/v1/models/sdk_operation_option_object_definition.go) (there's an associated `SDK Operation Option Object Definition Type` too)
* [Terraform Schema Object Definition](https://github.com/hashicorp/pandora/blob/main/tools/data-api-sdk/v1/models/terraform_schema_object_definition.go) (there's an associated `Terraform Schema Object Definition Type` too)

PR #3140, opened 2023-10-11
add appplatform@2023-09-01-preview

ms-henglu:


config/resource-manager.hcl
```diff
@@ -33,7 +33,7 @@ service "applicationinsights" {
 }
 service "appplatform" {
   name      = "AppPlatform"
-  available = ["2022-09-01-preview", "2022-12-01", "2023-05-01-preview", "2023-07-01-preview"]
+  available = ["2022-09-01-preview", "2022-12-01", "2023-05-01-preview", "2023-07-01-preview", "2023-09-01-preview"]
```
tombuildsstuff, line 36: Since we're no longer using the older versions, can we remove these too

```suggestion
  available = ["2023-07-01-preview", "2023-09-01-preview"]
```

tombuildsstuff (changes requested):
One comment about cleaning up the older versions - but this otherwise LGTM 👍

tombuildsstuff (approved):

PR #3151, opened 2023-10-12
data-api: implement the methods to read in the new JSON api definitions

stephybun:
This PR adds the following:

- implements the methods `GetAll` and `GetByName` on the `ServicesRepository` to populate the `resourceManager` attribute which contains the loaded API definitions from the JSON files
- initialises the `ServicesRepository` for each source type (by source type I mean `microsoft-graph-beta`, `microsoft-graph-stable-v1` and `resource-manager`) and extends the server router to accept a `ServicesRespository`
- adds and threads through a `-services` flag to the `serve` command so the data api can be spun up for specific service/s only

TODO in the next PR:

- flesh out `ProcessResourceDefinitions` to unmarshal the remaining definition types e.g. `Model-Something.json`, `ResourceId-Something.json` etc and map them into the `ServiceApiVersionResourceDetails` struct
- add the Terraform endpoints and the data api to read those definitions in, unmarshall and map them over into `ServiceDetails`
- implement the cache functionality
- implement the `serve-watch` command
- use `path.Join` for concatenating paths


tools/data-api/internal/endpoints/v1/routing.go
```diff
@@ -21,35 +21,44 @@ type Options struct {
 	UsesCommonTypes bool
 }
 
-func Router(router chi.Router, options Options) {
+type Api struct {
+	ServicesRepository repositories.ServicesRepository
+	Services           *[]string
```
tombuildsstuff, line 26: Two things:

1. These can become private fields rather than public?
2. Since this list is being parsed down to the repository, why do we need `Services` stored here too?

tools/data-api/internal/repositories/services.go
```diff
@@ -1,227 +1,411 @@
...
+	//}, nil
 }
 
 func (s *ServicesRepositoryImpl) GetByName(serviceName string, serviceType ServiceType) (*ServiceDetails, error) {
-	allServices, err := s.GetAll(serviceType)
+	// GetByName builds the ServiceDetails for a singular service by calling processing functions to build the
+	// structs for the ServiceApiVersionDetails and ServiceApiVersionResourceDetails
+	serviceDirectory := fmt.Sprintf("%s/%s", s.directory, serviceName)
+	if _, err := os.Stat(serviceDirectory); os.IsNotExist(err) {
+		return nil, fmt.Errorf("service %s does not exist: %+v", serviceName, err)
```
tombuildsstuff, line 284: `%q` would be helpful here when there's whitespace (due to a typo/incorrect split) or similar

```suggestion
		return nil, fmt.Errorf("service %q does not exist: %+v", serviceName, err)
```

tools/data-api/main.go
```diff
@@ -9,14 +9,18 @@ import (
...
 	c := cli.NewCLI("data-api", "1.0.0")
 	c.Args = os.Args[1:]
 	c.Commands = map[string]cli.CommandFactory{
-		"serve": func() (cli.Command, error) {
+		"serve": commands.NewServeCommand(definitionsDirectory),
+		// TODO hook this up
+		"serve-watch": func() (cli.Command, error) {
 			return commands.ServeCommand{
 				Log: logger,
 			}, nil
```
tombuildsstuff, line 26: Probably better to be explicit this isn't supported yet?

```suggestion
			return nil, fmt.Error("TODO: implement me")
```

tools/data-api/internal/repositories/services.go
```diff
@@ -1,227 +1,411 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("processing version definitions for %s: %+v", version, err)
+			}
+			versionDefinitions[version] = versionDetails
+		}
+	}
+
+	return &ServiceDetails{
+		Name:             serviceName,
+		ResourceProvider: fmt.Sprintf("Microsoft.%s", serviceName),
```
tombuildsstuff, line 328: Resource Providers don't have to be within the `Microsoft.` namespace - we've already got some which aren't (`Nginx.Nginx` etc) - so this'd need to be stored alongside the data

In addition, the Service name that we're using can differ (e.g. `VMware` [pandora] vs `Avs` [azure-rest-api-specs]) so that it's clearer as to what this is - therefore these two are intentionally different

tools/data-api/internal/repositories/services.go
```diff
@@ -1,227 +1,411 @@
...
+	s.resourceManager = nil
+
+	return nil
+}
+
+func (s *ServicesRepositoryImpl) GetAll(serviceType ServiceType) (*[]ServiceDetails, error) {
+	// GetAll calls GetByName for all the service names passed to the serve command, or for all the services available in the api definitions directory
+	serviceDetails := make([]ServiceDetails, 0)
+
+	if s.resourceManager == nil {
```
tombuildsstuff, line 56: for a follow up PR, but we're gonna want to add a lock when reloading this data/clearing the cache so that we only load this a single time - so we should add TODOs here

tools/data-api/internal/repositories/services.go
```diff
@@ -1,227 +1,411 @@
...
+
+	// graphV2 is a map of service name to string
+	graphV2 *map[string]string
+
+	// resourceManager is map of service name to ServiceDetails, the ServiceDetails contain all the
+	// Service, Version and Resource definitions loaded and unmarshalled from the JSON API definitions
+	resourceManager *map[string]ServiceDetails
+
+	// serviceNames is a list of services which the server will load
+	serviceNames *[]string
```
tombuildsstuff, line 33: We should call out this is only intended for dev purposes:

```suggestion
	// serviceNames is a list containing the names of services which should be loaded
	// this allows the loading/parsing of a subset of services for faster iterations during development
	serviceNames *[]string
```

tools/data-api/internal/repositories/services.go
```diff
@@ -1,227 +1,411 @@
...
+}
+
+func NewServicesRepository(directory string, serviceType ServiceType, serviceNames *[]string) ServicesRepository {
+	return &ServicesRepositoryImpl{
+		directory:    fmt.Sprintf("%s%s", directory, serviceType),
```
tombuildsstuff, line 38: another one for `path.Join`

tools/data-api/internal/repositories/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+// Model-KeyVaultProperties.json -> type = Model and name = KeyVaultProperties
+func getDefinitionInfo(fileName string) (string, string) {
+	splitName := strings.Split(fileName, "-")
+
+	definitionType := splitName[0]
+	definitionName := strings.Split(splitName[1], ".")[0]
+
+	return definitionType, definitionName
+
+}
```
tombuildsstuff, line 53: This method assumes that we're only parsing JSON files - so `.DS_Store` files and similar (output by macOS Finder/Windows Explorer) would crash this, it's probably worth returning an error here if the file extension isn't `*.json` since that would indicate we're loading files we're not intending too?

tools/data-api/internal/repositories/services.go
```diff
@@ -1,227 +1,411 @@
...
+
+	// graphV1 is a map of service name to string
+	graphV1 *map[string]string
+
+	// graphV2 is a map of service name to string
+	graphV2 *map[string]string
+
+	// resourceManager is map of service name to ServiceDetails, the ServiceDetails contain all the
+	// Service, Version and Resource definitions loaded and unmarshalled from the JSON API definitions
+	resourceManager *map[string]ServiceDetails
```
tombuildsstuff, line 30: since each `api` gets it's own repository, we're going to have 3 separate instances of the `repository`, which should all be using the same format on-disk, so this can be reduced to:

```suggestion
	// services is a  map of Service Name to ServiceDetails
	// ServiceDetails contains all the Service, Version and Resource definitions which
	// are loaded and unmarshalled from the JSON API definitions
	services *map[string]ServiceDetails
```

tombuildsstuff (approved):
Left a few minor comments inline, but this otherwise LGTM 👍

PR #3152, opened 2023-10-12
Completeness changes / generation preparation for `DevCenter`

tombuildsstuff:
This PR fixes several issues related to the `DevCenter` Service, notably:

1. Introduces a placeholder Terraform Configuration for Dev Center, allowing the Dev Center and Dev Center Project resources to be generated once this PR has rolled out / https://github.com/hashicorp/terraform-provider-azurerm/pull/23530 has been merged
2. Adds support to `importer-rest-api-specs` for Dev Center being a dependency 
3. Implements a Swagger workaround for https://github.com/Azure/azure-rest-api-specs/pull/26189
4. Fixes a bug in `generator-terraform` where Optional SystemAndUserAssignedMap identities wouldn't get assigned correctly
5. Fixes a bug in `importer-rest-api-specs` where a field name would be solely truncated to `Uri` - and whilst I investigated making this `ServiceUri` conflicting input from other packages means when this is truncated to just `Uri`, we should use the original name.
6. Makes the `DevCenterID` field required for the `Dev Center Project` resource to workaround an API design issue where the `DevCenterID` is defaulted/implied - meaning that the resource could be created with the wrong `DevCenterID` - making this Required solves this issue.


tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_devcenter_26189.go
```diff
@@ -0,0 +1,77 @@
+package dataworkarounds
+
+import (
+	"fmt"
+	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
+)
+
+// This workaround fixes an issue in DevCenter where the field `DevCenterUri` is not
+// marked as ReadOnly, meaning that it's surfaced as an Optional field rather than
+// being Computed.
```
stephybun, line 10: You're also setting the DevCenterId field to Required here as well, could you add that in the description here?

stephybun (approved):
One minor comment about the workaround description, otherwise LGTM 👍

tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_devcenter_26189.go
tombuildsstuff: Good shout, will fix that now.

PR #3167, opened 2023-10-13
`tools/generator-go-sdk`: older versions of `automation` now use the new base layer

tombuildsstuff:


tools/generator-go-sdk/main.go
```diff
@@ -56,12 +56,6 @@ func main() {
 		"Subscription",
 		"TimeSeriesInsights",
 
-		// Automation @ 2022-08-08 uses the new base layer, so let's invert the older versions for now
-		"Automation@2015-10-31",
-		"Automation@2019-06-01",
-		"Automation@2020-01-13-preview",
-		"Automation@2021-06-22",
```
tombuildsstuff, line 63: Checking the Provider - this one's no longer present in the provider (so is safe to remove today) - but the other 3 will require code changes

mbfrahry (approved):
LGTM!

PR #3189, opened 2023-10-17
`tools/importer-rest-api-specs`: refactoring the importer logic

tombuildsstuff:
This PR refactors the importer logic to simplify the generation, by first parsing and processing the data as needed - before passing off to both of the the API Definition Generators.

This PR also renames the API Definition Generators from C# and JSON to `V1` and `V2` - and adds feature flags for both, so that we have a migration path.

The V2 API Definitions still require additional changes (including better support for the ObjectDefinition type), however this seemed like a good stopping point - but this removes the unnecessary `Namespace` logic from the `V2` importer and then threads through a bunch of TODOs for future work.


tools/importer-rest-api-specs/pipeline/run_importer.go
```diff
@@ -135,24 +138,22 @@ func runImportForService(input RunInput, serviceName string, apiVersionsForServi
 			return fmt.Errorf(fmt.Sprintf("generating Terraform Example Usage for Service %q / Version %q: %+v", serviceName, apiVersion, err))
 		}
 
-		versionLogger.Trace("Task: Applying Overrides from Existing Data..")
```
mbfrahry, line 138: Are we doing anything with overrides?

mbfrahry (approved):
LGTM with a minor question about overrides

PR #3190, opened 2023-10-17
`tools/importer-rest-api-specs`: refactoring the common models within the `dataapigeneratorjson` package into a `models` directory

tombuildsstuff:
For now this package is going to be intentionally duplicated, as such splitting this out makes that easier to keep in sync for now - but also allows us to validate and document each model as it's split out.


mbfrahry (approved):
LGTM!

PR #3195, opened 2023-10-18
`tools/importer-rest-api-specs`: refactoring the Operation models within the `dataapigeneratorjson` package into a `models` directory

mbfrahry:
I have a few todos to track down:

- [x] Options vs OptionsObject
- [x] Do we need to do anything with that todo? 
```
if operation.ResponseObject != nil {
		responseObject.ReferenceName = operation.ResponseObject.ReferenceName
		responseObject.Type = operationModels.ObjectDefinitionType(operation.ResponseObject.Type)

		// TODO
		//if operation.FieldContainingPaginationDetails == nil {
		//	code = append(code, fmt.Sprintf(`		public override Type? ResponseObject() => typeof(%[1]s);`, *responseOperationTypeName))
		//} else {
		//	code = append(code, fmt.Sprintf(`		public override Type NestedItemType() => typeof(%[1]s);`, *responseOperationTypeName))
		//}
	}
```

- [x] and this todo
```
if operation.FieldContainingPaginationDetails != nil {
		HTTPMethod = "List"
		// TODO is there a todo here?
		// since this is a List operation we need to additionally output the HttpMethod
		// since it's possible that these are non-standard (e.g. AppConfig 2020-06-01 ListKeys
		// is a POST not a GET for a List operation)
		//if !strings.EqualFold(operation.Method, "GET") {
		//	method := dotNetNameForHttpMethod(operation.Method)
		//	code = append(code, fmt.Sprintf(`		public override System.Net.Http.HttpMethod Method() => System.Net.Http.%s;`, method))
		//}
	}
```

### DIFF
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
@@ -54,7 +54,7 @@ const (
 	// StringObjectDefinitionType signifies that this field contains a String.
 	StringObjectDefinitionType ObjectDefinitionType = "String"
 
-	// CsvObjectDefinitionType signifies that this field contains a CSV of simple types e.g. String, Integer, Float
+	// CsvObjectDefinitionType signifies that this field contains a CSV of simple types e.g. String, Integer, Float.
 	CsvObjectDefinitionType ObjectDefinitionType = "Csv"
 
 	// DictionaryObjectDefinitionType signifies that this field contains a Dictionary, the Dictionary's Value Type is defined as Nested Object Definition.
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
@@ -1,7 +1,5 @@
 package models
 
-import "github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
-
 // ObjectDefinition specifies additional information about a specific Object and any associated nested Objects
 type ObjectDefinition struct {
 	// ObjectDefinitionType defines what kind of ObjectDefinition this is, such as a Reference, String or List
@@ -87,17 +85,3 @@ const (
 	// TODO: others in the future https://github.com/hashicorp/pandora/issues/8 e.g.
 	// RFC3339NanoDateFormat DateFormat = "RFC3339Nano"
 )
-
-func ObjectDefinitionFromSchemaFieldFromImporterRestApiSpecs(input *models.ObjectDefinition) *ObjectDefinition {
-	if input == nil {
-		return nil
-	}
-
-	objectDefinition := ObjectDefinition{
-		ReferenceName: input.ReferenceName,
-		Type:          ObjectDefinitionType(input.Type),
-		NestedItem:    ObjectDefinitionFromSchemaFieldFromImporterRestApiSpecs(input.NestedItem),
-	}
-
-	return &objectDefinition
-}
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
@@ -1,12 +1,16 @@
 package models
 
-// NOTE: these types are intentionally undocumented atm, these'll be added in a follow-up PR
+import (
+	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
+)
 
+// ObjectDefinition specifies additional information about a specific Object and any associated nested Objects
 type ObjectDefinition struct {
 	Type ObjectDefinitionType `json:"type"`
 
 	ReferenceName *string `json:"referenceName"` // NOTE: we could split this into ConstantName and ModelName in time, but not today.
 
+	// NestedItem is a nested ObjectDefinition when Type is a Dictionary or List
 	NestedItem *ObjectDefinition `json:"nestedItem,omitempty"`
 
 	// NOTE: these 3 fields were previously located against the Field but instead should be located against the Object Definition
@@ -14,6 +18,20 @@ type ObjectDefinition struct {
 	MaxItems   *int        `json:"maxItems,omitempty"`
 	MinItems   *int        `json:"minItems,omitempty"`
 	DateFormat *DateFormat `json:"dateFormat,omitempty"`
+}
+
+func ObjectDefinitionFromSchemaFieldFromImporterRestApiSpecs(input *models.ObjectDefinition) *ObjectDefinition {
+	if input == nil {
+		return nil
+	}
+
+	objectDefinition := ObjectDefinition{
+		ReferenceName: input.ReferenceName,
+		Type:          ObjectDefinitionType(input.Type),
+		NestedItem:    ObjectDefinitionFromSchemaFieldFromImporterRestApiSpecs(input.NestedItem),
+	}
+
+	return &objectDefinition
 }
 
 type ObjectDefinitionType string
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
@@ -30,19 +30,38 @@ type ObjectDefinition struct {
 type ObjectDefinitionType string
 
 const (
-	// BooleanObjectDefinitionType is used to signify that this type is a simple Boolean.
-	BooleanObjectDefinitionType   ObjectDefinitionType = "Boolean"
-	DateTimeObjectDefinitionType  ObjectDefinitionType = "DateTime"
-	IntegerObjectDefinitionType   ObjectDefinitionType = "Integer"
-	FloatObjectDefinitionType     ObjectDefinitionType = "Float"
-	RawFileObjectDefinitionType   ObjectDefinitionType = "RawFile"
-	RawObjectObjectDefinitionType ObjectDefinitionType = "RawObject"
-	ReferenceObjectDefinitionType ObjectDefinitionType = "Reference"
-	StringObjectDefinitionType    ObjectDefinitionType = "String"
+	// BooleanObjectDefinitionType signifies that this type is a simple Boolean.
+	BooleanObjectDefinitionType ObjectDefinitionType = "Boolean"
 
-	CsvObjectDefinitionType        ObjectDefinitionType = "Csv"
+	// DateTimeObjectDefinitionType signifies that this field contains a DateTime value.
+	DateTimeObjectDefinitionType ObjectDefinitionType = "DateTime"
+
+	// IntegerObjectDefinitionType signifies that this field contains an Integer.
+	IntegerObjectDefinitionType ObjectDefinitionType = "Integer"
+
+	// FloatObjectDefinitionType signifies that this field contains a Float.
+	FloatObjectDefinitionType ObjectDefinitionType = "Float"
+
+	// RawFileObjectDefinitionType signifies that this field contains a byte value e.g. []byte.
+	RawFileObjectDefinitionType ObjectDefinitionType = "RawFile"
+
+	// RawObjectObjectDefinitionType signifies that this field contains an interface value e.g. interface{}.
+	RawObjectObjectDefinitionType ObjectDefinitionType = "RawObject"
+
+	// ReferenceObjectDefinitionType signifies that this field points to a Constant or a Model.
+	ReferenceObjectDefinitionType ObjectDefinitionType = "Reference"
+
+	// StringObjectDefinitionType signifies that this field contains a String.
+	StringObjectDefinitionType ObjectDefinitionType = "String"
+
+	// CsvObjectDefinitionType signifies that this field contains a CSV of simple types e.g. String, Integer, Float
+	CsvObjectDefinitionType ObjectDefinitionType = "Csv"
+
+	// DictionaryObjectDefinitionType signifies that this field contains a Dictionary, the Dictionary's Value Type is defined as Nested Object Definition.
 	DictionaryObjectDefinitionType ObjectDefinitionType = "Dictionary"
-	ListObjectDefinitionType       ObjectDefinitionType = "List"
+
+	// ListObjectDefinitionType signifies that this field contains a List, the List's Value Type is defined as a Nested Object Definition.
+	ListObjectDefinitionType ObjectDefinitionType = "List"
 
 	EdgeZoneObjectDefinitionType                                ObjectDefinitionType = "EdgeZone"
 	LocationObjectDefinitionType                                ObjectDefinitionType = "Location"
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
@@ -1,37 +1,30 @@
 package models
 
-import (
-	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
-)
+import "github.com/hashicorp/pandora/tools/importer-rest-api-specs/models"
 
 // ObjectDefinition specifies additional information about a specific Object and any associated nested Objects
 type ObjectDefinition struct {
+	// ObjectDefinitionType defines what kind of ObjectDefinition this is, such as a Reference, String or List
 	Type ObjectDefinitionType `json:"type"`
 
-	ReferenceName *string `json:"referenceName"` // NOTE: we could split this into ConstantName and ModelName in time, but not today.
+	// NOTE: we could split this into ConstantName and ModelName in time, but not today.
+	// ReferenceName is the name of the Constant or Model that this is a reference to
+	ReferenceName *string `json:"referenceName"`
 
 	// NestedItem is a nested ObjectDefinition when Type is a Dictionary or List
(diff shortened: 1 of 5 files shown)


tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
```diff
@@ -0,0 +1,73 @@
...
+	FieldContainingPaginationDetails string `json:"FieldContainingPaginationDetails,omitempty"`
+
+	// LongRunning specifies if this is a Long Running Operation, meaning that Clients
+	// should follow any `Location` headers to track the result of this operation
+	LongRunning bool `json:"LongRunning,omitempty"`
+
+	// HTTPMethod is the Method used for this operation, (e.g. `GET`, `POST`)
+	HTTPMethod string `json:"HTTPMethod,omitempty"`
+
+	// todo do we need this OptionsObject as well as Options?
```
manicminer, line 25: Looks to me like we don't need it? `OptionsObject.Name` seems to be the same as the operation name so perhaps it's redundant?

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -78,10 +50,10 @@ func codeForOperation(operationName string, operation models.OperationDetails, r
...
 	}
 
-	responseObject := ObjectDefinition{}
+	responseObject := operationModels.ObjectDefinition{}
 	if operation.ResponseObject != nil {
 		responseObject.ReferenceName = operation.ResponseObject.ReferenceName
-		responseObject.Type = ObjectDefinitionType(operation.ResponseObject.Type)
+		responseObject.Type = operationModels.ObjectDefinitionType(operation.ResponseObject.Type)
 
 		// TODO
```
manicminer, line 58: Theoretically these are different things but in practice from what I've seen we do the same with the result. Perhaps instead it'd be better to indicate whether the response is a single object or an array of objects? (Noting that paged and an array are not necessarily the same - not all endpoints may support paging)

manicminer (approved):
Thanks @mbfrahry, LGTM!

Couple of comments inline, but on the method override we'll definitely need to keep/represent this, it's used in a lot of services and I know a couple places in msgraph where this'll be needed.

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
tombuildsstuff: Agreed, the `Options` field should be sufficient here

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// RequestObject specifies the optional ObjectDefinition to be specified in the Request
+	RequestObject ObjectDefinition `json:"RequestObject,omitempty"`
+
+	// RequestObject specifies the optional ObjectDefinition to be returned by the Request
+	ResponseObject ObjectDefinition `json:"ResponseObject,omitempty"`
+
+	// UriSuffix specifies the suffix which should be appended to the ResourceID for this operation
+	// (e.g. `/shutdown`)
+	UriSuffix string `json:"UriSuffix,omitempty"`
```
tombuildsstuff, line 43: all 4 of these can be optional - so should be pointers:

```suggestion
	ResourceId *string `json:"ResourceId,omitempty"`

	// RequestObject specifies the optional ObjectDefinition to be specified in the Request
	RequestObject *ObjectDefinition `json:"RequestObject,omitempty"`

	// RequestObject specifies the optional ObjectDefinition to be returned by the Request
	ResponseObject *ObjectDefinition `json:"ResponseObject,omitempty"`

	// UriSuffix specifies the suffix which should be appended to the ResourceID for this operation
	// (e.g. `/shutdown`)
	UriSuffix *string `json:"UriSuffix,omitempty"`
```

`ResourceId` likely should become `ResourceIdName` so that it's clearer this is the name and not the URI etc?

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
```diff
@@ -0,0 +1,73 @@
...
+	// FieldContainingPaginationDetails is a reference to the field within the Response
+	// which contains the pagination details, (e.g. `nextLink`)
+	FieldContainingPaginationDetails string `json:"FieldContainingPaginationDetails,omitempty"`
+
+	// LongRunning specifies if this is a Long Running Operation, meaning that Clients
+	// should follow any `Location` headers to track the result of this operation
+	LongRunning bool `json:"LongRunning,omitempty"`
+
+	// HTTPMethod is the Method used for this operation, (e.g. `GET`, `POST`)
+	HTTPMethod string `json:"HTTPMethod,omitempty"`
```
tombuildsstuff, line 23: both of these are required, so there's no need for these to be `omitempty`?

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// ContentType specifies the format of the information being sent with the Operation (e.g. `application/json; charset=utf-8`)
+	ContentType string `json:"ContentType,omitempty"`
+
+	// ExpectedStatusCodes specifies is a list of Status Codes which are expected to be returned (e.g. 200, 201)
+	ExpectedStatusCodes []int `json:"ExpectedStatusCodes,omitempty"`
+
+	// FieldContainingPaginationDetails is a reference to the field within the Response
+	// which contains the pagination details, (e.g. `nextLink`)
+	FieldContainingPaginationDetails string `json:"FieldContainingPaginationDetails,omitempty"`
```
tombuildsstuff, line 16: This is optional/should be a pointer

```suggestion
	FieldContainingPaginationDetails *string `json:"FieldContainingPaginationDetails,omitempty"`
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
```diff
@@ -0,0 +1,73 @@
...
+type Operation struct {
+	// Name specifies the Display Name for this Operation (e.g. `Create`) which is also
+	// valid as an identifier.
+	Name string `json:"Name"`
+
+	// ContentType specifies the format of the information being sent with the Operation (e.g. `application/json; charset=utf-8`)
+	ContentType string `json:"ContentType,omitempty"`
+
+	// ExpectedStatusCodes specifies is a list of Status Codes which are expected to be returned (e.g. 200, 201)
+	ExpectedStatusCodes []int `json:"ExpectedStatusCodes,omitempty"`
```
tombuildsstuff, line 12: both of these are required so shouldn't be `omitempty`?

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/api_types.go
```diff
@@ -0,0 +1,34 @@
...
+	LegacySystemAndUserAssignedIdentityListObjectDefinitionType ObjectDefinitionType = "LegacySystemAndUserAssignedIdentityList"
+	LegacySystemAndUserAssignedIdentityMapObjectDefinitionType  ObjectDefinitionType = "LegacySystemAndUserAssignedIdentityMap"
+	SystemOrUserAssignedIdentityListObjectDefinitionType        ObjectDefinitionType = "SystemOrUserAssignedIdentityList"
+	SystemOrUserAssignedIdentityMapObjectDefinitionType         ObjectDefinitionType = "SystemOrUserAssignedIdentityMap"
+	UserAssignedIdentityListObjectDefinitionType                ObjectDefinitionType = "UserAssignedIdentityList"
+	UserAssignedIdentityMapObjectDefinitionType                 ObjectDefinitionType = "UserAssignedIdentityMap"
+	TagsObjectDefinitionType                                    ObjectDefinitionType = "Tags"
+	SystemDataObjectDefinitionType                              ObjectDefinitionType = "SystemData"
+	ZoneObjectDefinitionType                                    ObjectDefinitionType = "Zone"
+	ZonesObjectDefinitionType                                   ObjectDefinitionType = "Zones"
```
tombuildsstuff, line 33: we should document these

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
```diff
@@ -0,0 +1,12 @@
+package models
+
+type ObjectDefinition struct {
```
tombuildsstuff, line 3: todo we should document these

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/object_definition.go
```diff
@@ -0,0 +1,12 @@
...
+	// Type specifies what kind of Object Definition this is (a Boolean, UserAssignedIdentityMap, etc).
+	Type ObjectDefinitionType `json:"Type"`
+
+	// ReferenceName specifies the name of the Constant or Model that this is a reference to (ie ConfigurationStoreProperties, EncryptionProperties)
+	ReferenceName *string `json:"ReferenceName,omitempty"`
+
+	// NestedItem is a nested ObjectDefinition when Type is a Dictionary or List
+	NestedItem *ObjectDefinition `json:"ObjectDefinition,omitempty"`
```
tombuildsstuff, line 11: the json name should match:

```suggestion
	NestedItem *ObjectDefinition `json:"NestedItem,omitempty"`
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
```diff
@@ -0,0 +1,73 @@
...
+	QueryString *string `json:"QueryString,omitempty"`
+
+	// Required specifies whether this Option must be specified in the Request
+	Required bool `json:"Required"`
+
+	// Field specifies the DisplayName of the Option (e.g. `ConstantOption`, `SecondVal`)
+	Field string `json:"Field"`
+
+	// FieldType specifies what kind of Option this is (e.g. `String`, `Bool`)
+	FieldType string `json:"FieldType"`
```
tombuildsstuff, line 72: this needs to be a Constant (see `ConstantType` for an example)

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
tombuildsstuff: Both Requests and Responses can be infinitely scoped (in theory) e.g. a List of a List of a Dictionary of string:float values - but in addition, Request and Response are both optional - so we need to both handle both recursively - and handle them being nilable/not having a default value

Example: some DELETE operations have neither a request or a response model - they're simply a DELETE returning an LRO or not

tombuildsstuff (changes requested):
hey @mbfrahry 

Thanks for this PR - I've taken a look through and left some comments inline, but if we can fix those up this should otherwise be good to go 👍

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/api_types.go
mbfrahry: These constants feel straightforward enough to not need to be documented. I'm not sure what we would document here other than:

```
        // TagsObjectDefinitionType specifies an ObjectDefinition of type Tags
	TagsObjectDefinitionType                                    ObjectDefinitionType = "Tags"
```

Is that what you are looking for or something more specific?

tombuildsstuff: it's more the complex use-cases (e.g. Csv, Dictionary, List and Custom Types) which are worth calling out:

```
// BooleanObjectDefinitionType specifies that this field contains a Boolean.
//
// ListObjectDefinitionType specifies that this field contains a List, where the List's Value Type will be defined as a Nested Object Definition.
//
// ZonesObjectDefinitionType specifies that this field contains Zones.
```

However if we're doing a subset we might as well do the lot for consistency, hence the 3 examples above - WDYT?

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -70,14 +36,8 @@ func codeForOperation(operationName string, operation models.OperationDetails, r
...
-	if operation.RequestObject != nil {
-		requestObject.ReferenceName = operation.RequestObject.ReferenceName
-		requestObject.Type = ObjectDefinitionType(operation.RequestObject.Type)
-
-	} else if strings.EqualFold(operation.Method, "POST") || strings.EqualFold(operation.Method, "PUT") {
-		// Post and Put operations should have one but it's possible they don't
+	if operation.RequestObject == nil && (strings.EqualFold(operation.Method, "POST") || strings.EqualFold(operation.Method, "PUT")) {
+		// Post and Put operations should have a RequestObject one but it's possible they don't
 		// TODO what goes here?
 	}
```
tombuildsstuff, line 42: This logic can be removed - it originally made sense but unfortunately the API Definitions prove otherwise

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -155,19 +95,18 @@ func codeForOperation(operationName string, operation models.OperationDetails, r
...
 		HTTPMethod:                       HTTPMethod,
-		OptionsObject:                    optionsObject,
 		Options:                          options,
-		ResourceId:                       resourceId,
-		RequestObject:                    requestObject,
-		ResponseObject:                   responseObject,
-		UriSuffix:                        uriSuffix,
+		ResourceIdName:                   pointer.To(resourceId),
+		RequestObject:                    operationModels.ObjectDefinitionFromSchemaFieldFromImporterRestApiSpecs(operation.RequestObject),
+		ResponseObject:                   operationModels.ObjectDefinitionFromSchemaFieldFromImporterRestApiSpecs(operation.ResponseObject),
```
tombuildsstuff, line 108: (as above) can we use `mapObjectDefinition` (from this package) for these two?

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -86,24 +46,8 @@ func codeForOperation(operationName string, operation models.OperationDetails, r
...
-		//}
-	}
-
-	optionsObject := OptionsObject{}
-	options := make([]Option, 0)
+	options := make([]operationModels.Option, 0)
 	if len(operation.Options) > 0 {
-		optionsObject.Name = operationName
-
 		sortedOptionsKeys := make([]string, 0)
```
tombuildsstuff, line 51: the `options` assignment above will want moving inside of here:

```suggestion
		options :=make([]dataApiModels.Option, 0)
		sortedOptionsKeys := make([]string, 0)
```

since we want to conditionally assign this, as not all operations have Options

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/operations.go
tombuildsstuff: Yep, these can be a subset of ObjectDefinition (e.g. a Bool/Float/Int/String or a CSV of any of those) - we should however build this as a new type of ObjectDefinition (e.g. `OptionsObjectDefinition`) since these values will always be distinct from a regular OptionsDefinition

tombuildsstuff (changes requested):
hey @stephybun @mbfrahry 

Thanks for pushing those changes - I've taken a look though and left some comments inline, but on the whole this is looking pretty good - the main thing is that we need to create an ObjectDefinition for Options, presumable `OptionsObjectDefinition` which supports the subset of types that that supports - but otherwise this is looking pretty good, nice one 👍 

Thanks!

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -59,51 +26,49 @@ func codeForOperation(operationName string, operation models.OperationDetails, r
...
-	responseObject := ObjectDefinition{}
-	if operation.ResponseObject != nil {
-		responseObject.ReferenceName = operation.ResponseObject.ReferenceName
-		responseObject.Type = ObjectDefinitionType(operation.ResponseObject.Type)
+	uriSuffix := ""
+	if operation.UriSuffix != nil {
+		uriSuffix = *operation.UriSuffix
+	}
+
+	HTTPMethod := strings.Title(strings.ToLower(operation.Method))
```
tombuildsstuff, line 45: There's enough assumptions throughout the codebase which assume this is upper-cased, we should do that here - although tbh we can inline this below, so:

```suggestion
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -59,51 +26,49 @@ func codeForOperation(operationName string, operation models.OperationDetails, r
...
-	options := make([]Option, 0)
-	if len(operation.Options) > 0 {
-		optionsObject.Name = operationName
+	op := dataApiModels.Operation{
+		Name:                             operationName,
+		ContentType:                      contentType,
+		ExpectedStatusCodes:              statusCodes,
+		FieldContainingPaginationDetails: operation.FieldContainingPaginationDetails,
+		LongRunning:                      longRunning,
+		HTTPMethod:                       HTTPMethod,
```
tombuildsstuff, line 63:
```suggestion
		HTTPMethod:                       strings.ToUpper(operation.Method),
```

tombuildsstuff (approved):
A few minor things but otherwise LGTM 👍

PR #3214, opened 2023-10-20
[WIP] Process the Terraform piece of dataapiv2

mbfrahry:
This PR starts the process of mapping the Terraform information into the API. There are still pieces missing and we're moving models into a separate package so this is just a scaffold to build on top of


mbfrahry:
Closing in favor of #3268

PR #3215, opened 2023-10-23
tools/importer-rest-api-specs: updating the Resource ID models

tombuildsstuff:
This PR splits out the Resource ID model, adds missing fields (Constant Name) and documents each of the fields/expected values.

Notably this PR updates the Resource ID types to match those we're expecting in the Data API - with the `Value` field for a Resource ID segment is used as the literal/fixed/required value - rather than being reused for a Constant - which means this'll be nil/unset for other scenarios (as we're expecting).


stephybun (approved):
Thanks @tombuildsstuff LGTM 👍

PR #3223, opened 2023-10-24
`tools/importer-rest-api-specs`: refactoring the V2 API Definitions for Models/Fields/Object Definitions

tombuildsstuff:
This PR refactors the V2 API Definitions for Models/Fields/Object Definitions - such that:

* Updates the structs for Models, Field (now `ModelField`, to differentiate it from a Terraform Schema Field) and ObjectDefinition to fix issues outlined in https://github.com/hashicorp/pandora/pull/3185 (and spotted whilst passing through0
* A working directory is assumed to exist at `./api-definitions` - as such `importer-rest-api-specs` can rely on this existing and generate a `resource-manager` directory within it.
* Removes the `namespace` fields used within V2 of the API Definitions (since this is specific to V1)

---

There's one change to V2 of the API Definitions which doesn't exist in V1 worth mentioning, which is that the fields `MaxItems`, `MinItems` and `DateFormat` are now housed within the ObjectDefinition and not against the field. Neither `MaxItems`, `MinItems` (or the other, unmapped field, `UniqueItems`) are currently used in the Go SDK - these exist purely for the Terraform generation functionality to determine how to build up the Schema in the case of a Csv/Dictionary/List.

As such when we come to expose this in Data API V2, when parsing the data we'll need to account for this in the short-term (so that we can keep this working in the current schema/structure being returned from the API) and then (in the future) output this within the `ObjectDefinition` types so that we can migrate this across. It'd be worth adding an issue for that, but I figured it'd be worth calling that out here first.

### DIFF
--- a/tools/importer-rest-api-specs/pipeline/run_importer.go
+++ b/tools/importer-rest-api-specs/pipeline/run_importer.go
@@ -48,7 +48,7 @@ func runImporter(input RunInput, generationData []discovery.ServiceInput, swagge
 		}
 
 		if featureflags.GenerateV2APIDefinitions {
-			serviceDirectoryYaml := fmt.Sprintf("%s%s/%s", input.OutputDirectoryJson, "resource-manager", serviceName)
+			serviceDirectoryYaml := fmt.Sprintf("%s/%s", input.OutputDirectoryJson, serviceName)
 			logger.Debug("recreating JSON directory %q for Service %q", serviceDirectoryYaml, serviceName)
 			if err := dataapigenerator.RecreateDirectory(serviceDirectoryYaml, logger); err != nil {
 				fmt.Errorf("recreating JSON directory %q for service %q", serviceDirectoryYaml, serviceName)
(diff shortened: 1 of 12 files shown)


tools/importer-rest-api-specs/pipeline/run_importer.go
```diff
@@ -48,7 +48,7 @@ func runImporter(input RunInput, generationData []discovery.ServiceInput, swagge
 		}
 
 		if featureflags.GenerateV2APIDefinitions {
-			serviceDirectoryYaml := fmt.Sprintf("%s%s/%s", input.OutputDirectoryJson, "resource-manager", serviceName)
+			serviceDirectoryYaml := fmt.Sprintf("%s/%s", input.OutputDirectoryJson, serviceName)
```
stephybun, line 51:
```suggestion
			serviceDirectoryJson := fmt.Sprintf("%s/%s", input.OutputDirectoryJson, serviceName)
```

stephybun (approved):
One minor comment on the variable name otherwise LGTM

PR #3224, opened 2023-10-24
tools/importer-rest-api-specs: updating the v2 models to use camelCase

tombuildsstuff:
Given the conventions in JSON, we should conform to this - the side-effect is this gives us an easy tell for models which have been reviewed/refactored verses those which haven't.

This PR also splits out a struct for `MetaData`, since we'll want to reuse this for both ARM, MS Graph and handwritten data - and honestly could make sense for us to serve from the Data API (for example, allowing the new differ tool to output a `compare` link, to show differences in the source data) - so splitting this out into a dedicated struct seems worthwhile.


stephybun (approved):
Other than import ordering LGTM

PR #3235, opened 2023-10-25
tools/importer-rest-api-specs: updating the `metadata` file to include the SourceData and SourceInformation

tombuildsstuff:
This PR adds the fields `SourceData` and `SourceInformation` to the `metadata` file, allowing for multiple data sources (i.e. directories, `resource-manager` for generated and `handwritten-resource-manager` for handwritten) to provide Data for a single Service (e.g. AzureResourceManager/MicrosoftGraph) - so that the Data API can be configured to load this.

This also fixes a couple of other bugs spotted when generating the full data set:

* Operations - `ContentType` and `ExpectedStatusCodes` must always be output (the Go SDK generator assumes both are returned from the Data API, and these default values previously came from the base types in C#)
* Operations - `ResourceIdName` and `UriSuffix` are now pointers rather than empty strings, both generators rely on these being nil when omitted (since both can be specified together or individually)
* Models - outputting the `Name` field

### DIFF
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/api_version_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/models/api_version_definition.go
@@ -12,9 +12,6 @@ type ApiVersionDefinition struct {
 	// Generate specifies whether this API Version should be generated or not.
 	Generate bool `json:"generate"`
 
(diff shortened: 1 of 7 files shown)


stephybun (approved):
LGTM 🥧

PR #3238, opened 2023-10-25
Refactoring the dataapigeneratorjson Terraform Models into the `models` package

mbfrahry:


tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_definition.go
```diff
@@ -0,0 +1,56 @@
+package models
+
+// ResourceDefinition describes a Resource with information specific to Terraform
+type ResourceDefinition struct {
+	// DisplayName specifies the human-readable name for this Resource, used in the Documentation. (e.g. Load Test)
+	DisplayName string `json:"DisplayName"`
```
tombuildsstuff, line 6: for each of these we should be using `camelCase` for the json field names

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_definition.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// ResourceIdName specifies the name of the Resource ID type used for this Resource.
+	ResourceIdName string `json:"Id"`
```
tombuildsstuff, line 9: This one should match the field name:

```suggestion
	ResourceIdName string `json:"resourceIdName"`
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_tests.go
```diff
@@ -4,18 +4,12 @@ import (
 	"encoding/json"
 	"strings"
 
+	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/components/dataapigeneratorjson/models"
```
tombuildsstuff, line 7: for consistency (given there's an existing `models` package within `importer-rest-api-specs`) can we alias this import:

```suggestion
	dataApiModels "github.com/hashicorp/pandora/tools/importer-rest-api-specs/components/dataapigeneratorjson/models"
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_schema.go
```diff
@@ -111,14 +88,14 @@ func topLevelFieldObjectDefinition(input resourcemanager.TerraformSchemaFieldObj
 	return input
 }
 
-func objectDefinitionfromSchemaField(input resourcemanager.TerraformSchemaFieldObjectDefinition) *SchemaFieldObjectDefinition {
-	objectDefinition := SchemaFieldObjectDefinition{
+func objectDefinitionfromSchemaField(input resourcemanager.TerraformSchemaFieldObjectDefinition) *dataApiModels.ObjectDefinition {
+	objectDefinition := dataApiModels.ObjectDefinition{
 		ReferenceName: input.ReferenceName,
-		Type:          TerraformSchemaFieldType(input.Type),
+		Type:          dataApiModels.ObjectDefinitionType(input.Type),
```
tombuildsstuff, line 94: this assumes these are 1:1 compatible, when the backing values could be different - can we instead define a map for these and pull out the value (`dataApiModels.ObjectDefinitionType`) from the key (`resourcemanager.TerraformSchemaFieldObjectDefinitionType`)

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_mappings.go
```diff
@@ -3,40 +3,16 @@ package dataapigeneratorjson
 import (
 	"encoding/json"
 
+	"github.com/hashicorp/pandora/tools/importer-rest-api-specs/components/dataapigeneratorjson/models"
```
tombuildsstuff, line 6: for consistency (and since there's already a `models` package within `importer-rest-api-specs`) - could we import-alias this for now?

```suggestion
	dataApiModels "github.com/hashicorp/pandora/tools/importer-rest-api-specs/components/dataapigeneratorjson/models"
```

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_tests.go
```diff
@@ -0,0 +1,16 @@
+package models
+
+// TerraformResourceTestConfig describes the various test configurations that a Resource will be tested against
+type TerraformResourceTestConfig struct {
+	// Basic is the most basic hcl that only contains the minimum attributes needed to create this specific resource
+	Basic string `json:"Basic"`
```
tombuildsstuff, line 6: (as above) can we make these `camelCase`?

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_schema.go
```diff
@@ -0,0 +1,40 @@
...
+// SchemaModel holds information about all the fields associated with a Terraform Resource
+type SchemaModel struct {
+	// Fields describes all the fields associated with a Terraform Resource
+	Fields []SchemaField `json:"Fields"`
+}
+
+// SchemaField describes information about an attribute that will be mapped into the Terraform Schema of a resource
+type SchemaField struct {
+	// HclName specifies the snaked cased attribute name (e.g. resource_group_name)
+	HclName string `json:"HclName"`
```
tombuildsstuff, line 14: (as above) can we make these `camelCase`

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_schema.go
```diff
@@ -0,0 +1,40 @@
+package models
+
+import "github.com/hashicorp/pandora/tools/sdk/resourcemanager"
+
+// SchemaModel holds information about all the fields associated with a Terraform Resource
+type SchemaModel struct {
+	// Fields describes all the fields associated with a Terraform Resource
+	Fields []SchemaField `json:"Fields"`
```
tombuildsstuff, line 8: presumably this also needs the Name in here?

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_mappings.go
```diff
@@ -0,0 +1,51 @@
+package models
+
+type ResourceMapping struct {
+	// ResourceIdMapping is a list of Resource ID Mappings, defining how a top-level Schema Field
+	// gets mapped to/from a Resource ID Segment.
+	ResourceIdMapping *[]ResourceIdMapping `json:"ResourceIdMapping,omitempty"`
```
tombuildsstuff, line 6: (camelcasing as below)

tombuildsstuff (approved):
Left some comments I'm going to push a commit to fix, but this otherwise LGTM 👍

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_schema.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// SchemaField describes information about an attribute that will be mapped into the Terraform Schema of a resource
+type SchemaField struct {
+	// HclName specifies the snaked cased attribute name (e.g. resource_group_name)
+	HclName string `json:"HclName"`
+
+	// Documentation describes what this attribute is
+	Documentation *string `json:"Documentation,omitempty"`
```
tombuildsstuff, line 17: this should be a struct with a field `Markdown` inside it, to give us the means of adding other fields (e.g. a list of uris to examples) in here in the future

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_schema.go
```diff
@@ -0,0 +1,40 @@
...
+	Computed *bool `json:"Computed,omitempty"`
+
+	// ForceNew specifies whether a change in this attribute requires Terraform to recreate the resource
+	ForceNew *bool `json:"ForceNew,omitempty"`
+
+	// Name specifies the Display Name for this attribute
+	Name string `json:"Name"`
+
+	// Constants describes the specific values this attribute can be
+	Constants *resourcemanager.ConstantDetails `json:"Constants,omitempty"`
```
tombuildsstuff, line 35: digging into this one, this should be a `Validation` object, allowing us to specify the constant where the values for this field come from (which then allows us to generate a `Validate` function which uses those values)

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_schema.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// ObjectDefinition describes additional information about the specific type of attribute this is
+	// and any nested items associated with this attribute
+	ObjectDefinition *ObjectDefinition `json:"ObjectDefinition,omitempty"`
```
tombuildsstuff, line 39: I'd missed this one because of the inconsistent struct naming at first - but this needs to be a Terraform specific Object Definition - since the fields available in the Terraform Schema differ from what's available in the SDK (we may expose a `[]byte` field in the SDK as a JSON payload, for example)

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_schema.go
```diff
@@ -0,0 +1,40 @@
+package models
+
+import "github.com/hashicorp/pandora/tools/sdk/resourcemanager"
+
+// SchemaModel holds information about all the fields associated with a Terraform Resource
+type SchemaModel struct {
```
tombuildsstuff, line 6: all of the Terraform models want a `Terraform` prefix on these

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_resource_tests.go
```diff
@@ -0,0 +1,16 @@
...
+
+	// RequiresImport is the hcl that will be used to confirm a Resource is imported correctly
+	RequiresImport string `json:"RequiresImport"`
+
+	// CompleteConfig is the most complete hcl that contains all the attributes a Resource manages
+	CompleteConfig *string `json:"CompleteConfig,omitempty"`
+
+	// TemplateConfig is the hcl that contains the various resources needed to provision this specific resource
+	TemplateConfig *string `json:"TemplateConfig,omitempty"`
+}
```
tombuildsstuff, line 16: whilst we're not using it right now, the Terraform generator supports "other" tests (e.g. run these configs) - so we should also support this - as such we'll need to include a field for this one:

```suggestion
	OtherTests *map[string][]string `json:"otherTests,omitempty"`
}
```

tombuildsstuff (changes requested):
Spotted a few things that I'd missed the first time around having pulled this down locally

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_mappings.go
```diff
@@ -0,0 +1,189 @@
...
+
+		case resourcemanager.ModelToModelMappingDefinitionType:
+			{
+				// ModelToModel mappings need to be output both for Fields and for the Models themselves
+				// this is because a ModelToModel mapping must exist from the Schema Model to the SDK Model
+				// but also from a given Schema Field to a given SDK Model.
+				//
+				// This allows both the mapping function between two models to be generated (the `ModelToModelMappings`)
+				// and the mapping between a given Schema Field and an SDK Model (so that we know to call the
+				// mapping function defined in the `ModelToModelMappings`).
```
stephybun, line 41: It's unclear to me what this comment is really trying to say or add in this context of a case/switch and is potentially causing more confusion.

Is it saying that regardless of what the schema looks like there always needs to be a `ModelToModel` present with values in it? or empty? 
What happens if it's absent? 

To explain my confusion further:
`// This allows both the mapping function between two models to be generated (the ModelToModelMappings)`
* Both here indicating two mapping functions? 
* Allows meaning there's further logic down the line that we'll only run if a `ModelToModel` mapping exists?

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_mappings.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+
+		case dataApiModels.ManualTerraformFieldMappingDefinitionType:
+			{
+				key = fmt.Sprintf("%s-%s", string(item.Type), item.Manual.MethodName)
+			}
+
+		default:
+			{
+				// the likelihood of us getting here without being caught above it low, and so would
```
stephybun, line 150:
```suggestion
				// the likelihood of us getting here without being caught above is low, and so would
```

tools/importer-rest-api-specs/pipeline/task_generate_data_api_definitions_v2.go
```diff
@@ -20,7 +20,7 @@ func (pipelineTask) generateApiDefinitionsV2(serviceName string, apiVersions []m
 	// Generate each of the API Versions for this Service
 	for _, apiVersion := range apiVersions {
 		logger.Debug("Generating Data API Definition in JSON..")
-		yamlGenerator := dataapigeneratorjson.NewForApiVersion(apiVersion.ServiceName, apiVersion.ApiVersion, outputDirectory, swaggerGitSha, resourceProvider, terraformPackageName, logger.Named("Data API Generator"))
+		yamlGenerator := dataapigeneratorjson.NewForApiVersion(apiVersion.ServiceName, apiVersion.ApiVersion, outputDirectory, resourceProvider, terraformPackageName, logger.Named("Data API Generator"))
```
stephybun, line 23:
```suggestion
		v2Generator := dataapigeneratorjson.NewForApiVersion(apiVersion.ServiceName, apiVersion.ApiVersion, outputDirectory, resourceProvider, terraformPackageName, logger.Named("Data API Generator"))
```

stephybun (commented):
Thanks @tombuildsstuff.

Overall this looks fine but I think there questions/confusions around some of the comments/explanations in the documentation which should be addressed.

tools/importer-rest-api-specs/components/dataapigeneratorjson/generate_terraform_resource.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func (s Generator) generateTerraformResourceDefinition(resourceLabel string, details resourcemanager.TerraformResourceDetails) error {
+	s.logger.Trace("Building Definition for the Terraform Resource..")
+	resourceDefinition, err := buildTerraformResourceDefinition(resourceLabel, details)
+	if err != nil {
+		return fmt.Errorf("building Terraform Resource Definition: %+v", err)
+	}
+	s.logger.Trace("Marshalling Definition for the Terraform Resource..")
+	resourceDefinitionCode, err := json.MarshalIndent(resourceDefinition, "", "\t")
```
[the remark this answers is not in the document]
tombuildsstuff: Yeah agreed, tbh at this point we should be able to move the marshal function into the method which writes the file to disk - which both makes this consistent and reduces the TLOC

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_mappings.go
```diff
@@ -0,0 +1,131 @@
+package models
+
+// TerraformMappingDefinition defines a Mappings between a Terraform Resource and
+// the Models used within the SDK.
+// In the future this'll also be used for defining the mappings between a Terraform
+// Data Source and the Models - hence this not being specific to a Data Source.
```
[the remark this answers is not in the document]
tombuildsstuff: Yep, indeed - this should be Resource 🤦

tools/importer-rest-api-specs/pipeline/task_generate_data_api_definitions_v2.go
tombuildsstuff: Everything within this file is related to the V2 Generator, so on reflection I'm going to remove the version prefix altogether (and make this `apiVersionGenerator` and `serviceGenerator`)

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_schema_model.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+// TerraformSchemaModel defines a model used in the Terraform Schema, which is used to output
+// both a Struct and the Terraform Schema itself.
```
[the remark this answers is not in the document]
tombuildsstuff: Pushed an update for this better documenting this one

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_mappings.go
```diff
@@ -0,0 +1,131 @@
...
+type TerraformMappingDefinition struct {
+	// FieldMappings defines the mappings between Schema Fields within a Schema Model and the Sdk Fields
+	// within an SDK Model.
+	FieldMappings *[]TerraformFieldMappingDefinition `json:"fieldMappings,omitempty"`
+
+	/// ModelToModelMappings is a list of the Model to Model Mappings, used to generate the Mapping
+	// functions between Schema and SDK Models.
```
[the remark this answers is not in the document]
tombuildsstuff: both refer to a Schema Model:

1. a Field Mapping refers to a field within a Schema Model -> a field in the SDK
2. a ModelToModel Mapping refers to a model within the SchemaModel that gets mapped onto a Model in the SDK

The first generates the mappings between the fields (such as `a = b`), whereas the second generates the functions which houses those assignments (e.g. `func mapFooToBar(input Foo) Bar {`)

I'll update the description here to mention this a TerraformSchemaModel in both cases.

tools/importer-rest-api-specs/components/dataapigeneratorjson/models/terraform_schema_object_definition.go
```diff
@@ -0,0 +1,100 @@
...
+	// This exists when Type is set to ReferenceTerraformSchemaObjectDefinitionType.
+	ReferenceName *string `json:"referenceName"`
+
+	// Type specifies the Type that this represents, for example a String or a Location.
+	Type TerraformSchemaObjectDefinitionType `json:"type"`
+}
+
+// TerraformSchemaObjectDefinitionType specifies the Type of Object that is used in the Terraform Schema
+// NOTE: these are intentionally separate to the regular (SDK) Object Definition since these have different
+// types - and are not 1:1 compatible (for example, requiring an Expand/Flatten/Transform function).
```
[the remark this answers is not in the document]
tombuildsstuff: Missed this one, reworked the comment to avoid this

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_mappings.go
tombuildsstuff: A Field Mapping between a Terraform Schema Model and an SDK Model must have a matching `ModelToModel` (`TerraformModelToModelMappingDefinition`) mapping, so that we know to generate the Mapping functions between the Terraform Schema Model.

As such when we parse the `FieldModelToModel` mappings - that is "The Field X is a model which should be mapped to this other model" - we need to ensure that a `ModelToModel` mapping exists - so that we can output both the call to the mapping function (for the `FieldModelToModel` model) and the mapping function itself (from the `ModelToModel` model).

If the `ModelToModel` mapping doesn't exist but the `FieldModelToModel` mapping does, we'll output a call to a function which doesn't exist - leading to a compilation failure. As such this logic exists both to capture that requirement (whilst we could instead raise an error here, I suspect this'll become the source of truth for that going forwards).

Does that make sense / how's best to explain that one?

stephybun: I think that mostly makes sense, the second paragraph is a little unclear still, is the message still correct if the following adjustments (made in bold italic) are made

>As such when we parse the FieldModelToModel mappings - that is "The **_Field X in a model_** which should be mapped to this other model" - we need to ensure that a ModelToModel mapping exists - so that we can output both the call to the mapping function (for the FieldModelToModel model) and the _**model mapping**_ function itself (from the ModelToModel model).

I think I'm struggling to understand what a "Field which is a model" looks like? and how a field, which for me would be a singular flat property, can be mapped to a model which I view as something that contains multiple properties.

PR #3258, opened 2023-11-01
config: generating dev_center_galleries

jiaweitao001:
* Add support for `dev_center_galleries`


stephybun:
Thanks for this PR @jiaweitao001.

We currently don't have any automation around validating generated resources when they get added to Pandora (something we're working on). Could you please generate the resource locally and copy the contents of all of the files  into a gist for us to look at?

PR #3266, opened 2023-11-01
docs: documenting the different types of Mappings

tombuildsstuff:
This PR adds a new `internals` piece of documentation covering Mappings, used to map the Terraform Schema onto the Types used in the SDK and back again.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,6 +34,7 @@ Specific information on each tool can be found in the README for each tool, [for
 * [How to add a new Common ID](common-ids.md).
 * [Resource Manager: How to import a new API Version or Service into Pandora](resource-manager-service-import.md).
 * [Resource Manager: Generating a new Resource](resource-manager-generate-new-resource.md).
+* [Internals: Types of (Terraform) Mappings](internals-mappings.md)
 
 ### Resource ID's as a primary concept
 
--- a/docs/internals-mappings.md
+++ b/docs/internals-mappings.md
@@ -1,0 +1,428 @@
+# Internals: Mappings
+
+Each Pandora-generated Terraform Resource is associated with a single version of the API - and the Terraform Schema Model(s) for this Resource are based on the API Version.
+
+Mappings define how to map each Field within each Terraform Schema Model onto each Field within the associated SDK Models (located in `hashicorp/go-azure-sdk`).
+
+> **Note:** Mappings are bidirectional - meaning we're mapping the Terraform Schema Field onto an SDK Field (in Terraform's Create/Update functions) and the SDK Field onto the Terraform Schema Field (in Terraform's Read function).
+
+There are three different types of Mappings:
+
+1. Field Mappings.
+2. Model to Model Mappings.
+3. Resource ID (Segment) Mappings.
+
+In each case, these map the fields in the Terraform Schema onto either the SDK Models - or the Resource ID for a given Resource.
+
+Whilst Mappings define how to map a Type of Field from the Schema to another Field, mapping the value of those fields is left to the Terraform Generator, for example where one field is Required and another is Optional - the Terraform Generator should handle passing either a reference/the value as needed.
+
+## 1: Field Mappings
+
+Each Field within a Terraform Schema Model has a Field Mapping which defines how to map the value from the Schema Field onto the associated SDK Field.
+
+The Type of Field Mapping that this is determines how the two values should be mapped onto one another - of which there are three possible Types:
+
+1. `DirectAssignment` - which specifies that these values should be directly assigned (e.g. `a = b`) onto one-another - doing any transformation necessary (e.g. `flatten` functions, handling one being Optional the other being Required etc).
+2. `ModelToModel` - which specifies that this Schema Field should be mapped onto another Model. This happens when the Fields `ObjectDefinition` is either a `Reference`, `Dictionary[string, Reference]`, `List[Reference]` or `Set[Reference]`.
+3. `Manual` - specifies that this Field must be manually mapped (for example due to complex behaviour).
+
+> _Future thought:_ additional Field Mapping types could be added to support additional use-cases. For example, transforming values (e.g. such as having an `Integer` used in the Terraform Schema, but a `String` exposed in the SDK - or to support handling `SkuName` etc). 
+
+### a. `DirectAssignment` Field Mappings
+
+A `DirectAssignment` mapping specifies that the value of a Field within the Terraform Schema should be mapped onto the value of a Field within the related (Go) SDK Model - and vica-versa.
+
+This requires that the `TerraformObjectDefinitionType` used in the Terraform Schema Field and the `ObjectDefinitionType` used in the Field within the SDK Model are compatible - for example mapping a `String` (Schema) onto a `String` (SDK).
+
+Mappings between the two different Object Definitions are defined within the Terraform Generator, both [for simple types](https://github.com/hashicorp/pandora/blob/d893d8c6533bbfc3feab00dea5fcf58f880897df/tools/generator-terraform/generator/mappings/assignment_direct.go#L10-L22) and [complex types requiring some form of transformation](https://github.com/hashicorp/pandora/blob/d893d8c6533bbfc3feab00dea5fcf58f880897df/tools/generator-terraform/generator/mappings/assignment_direct_transform.go#L5-L14).
+
+Conversions between Object Definition Types (for example the Schema containing an Integer but the SDK containing a String value) are not supported at this time - but if needed would likely be supported as a new type of Mapping.
+
+Direct Assignment Mappings support the following properties:
+
+* `schemaModelName` - specifies the name of the Terraform Schema Model where this value should be mapped from.
+* `schemaFieldPath` - specifies the path to the Field within the Terraform Schema Model defined in `schemaModelName` which this should be mapped from.
+* `sdkModelName` - specifies the name of the SDK Model where this value should be mapped onto.
+* `sdkFieldPath` - specifies the Path to the Field within the SdkModel where the Schema Field should be mapped onto.
+
+> **Note:** Whilst `schemaFieldPath` and `sdkFieldPath` are named `Path` to allow supporting indexed keys in the future (for example using `AppSettings.SomeSetting` to retrieve the value for the key `SomeSetting` within the Dictionary defined in the `AppSettings` property) - at this point in time these only support top-level fields (e.g. `Name`).
+
+An example of a Direct Assignment Mapping can be seen below:
+
+> **Note:** A more complete example can be seen under `Examples` below.
+
+```json
+{
+  "fieldMappings": [
+    {
+      "type": "DirectAssignment",
+      "directAssignment": {
+        "schemaModelName": "SolarSystemResourceSchema",
+        "schemaFieldPath": "Name",
+        "sdkModelName": "SolarSystem",
+        "sdkFieldPath": "Name"
+      }
+    }
+  ]
+}
+```
+
+### b. `ModelToModel` Field Mappings
+
+ModelToModel Field Mappings - not to be confused with ModelToModel Mappings (below) - allow mapping a Terraform Schema Model onto a field within an SDK Model.
+
+The best example of this is allowing the top-level Terraform Schema Model for a given Terraform Resource to map onto the `properties` object that exists in most payloads in Azure Resource Manager - allowing for fields which exist at the root of the Terraform Schema to be mapped to a nested object.
+
+Each ModelToModel Field Mapping must have an associated ModelToModel mapping, these are used together:
+
+* A ModelToModel Field Mapping is used to output the call to the mapping function itself.
+* A ModelToModel Mapping is used to output the mapping function, the fields which should be mapped are then inferred from the ModelToModel Field Mappings.
+
+> **Note:** A complete end-to-end example can be found below under `Example`.
+
+Model to Model Field Mappings support the following properties:
+
+* `schemaModelName` - specifies the name of the Terraform Schema Model that should be mapped onto the Field within the SDK Model.
+* `sdkModelName` - specifies the name of the SDK Model containing the SDK Field that the Terraform Schema Model should be mapped onto.
+* `sdkFieldPath` - specifies the Path to the Field within the SDK Model where the Terraform Schema Model should be mapped onto.
+
+An example of a Model to Model Field Mapping can be seen below: 
+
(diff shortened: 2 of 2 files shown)


docs/README.md
```diff
@@ -34,6 +34,7 @@ Specific information on each tool can be found in the README for each tool, [for
 * [How to add a new Common ID](common-ids.md).
 * [Resource Manager: How to import a new API Version or Service into Pandora](resource-manager-service-import.md).
 * [Resource Manager: Generating a new Resource](resource-manager-generate-new-resource.md).
+* [Internals: Types of (Terraform) Mappings](internals-mappings.md)
```
stephybun, line 37: I think it makes sense to have a clear distinction between information that is user facing like guides and information non-user facing, relating to the inner workings of Pandora.

So can we have a separate grouping on this page for things like architecture/technical details and move this there?

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+
+There are three different types of Mappings:
+
+1. Field Mappings.
+2. Model to Model Mappings.
+3. Resource ID (Segment) Mappings.
+
+In each case, these map the fields in the Terraform Schema onto either the SDK Models - or the Resource ID for a given Resource.
+
+Whilst Mappings define how to map a Type of Field from the Schema to another Field, mapping the value of those fields is left to the Terraform Generator, for example where one field is Required and another is Optional - the Terraform Generator should handle passing either a reference/the value as needed.
```
stephybun, line 17: What is meant by `type of Field`. It looks like the types are mentioned down below e.g. `DirectAssignment`, `ModelToModel` etc. but that's not clear here.

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+
+1. `DirectAssignment` - which specifies that these values should be directly assigned (e.g. `a = b`) onto one-another - doing any transformation necessary (e.g. `flatten` functions, handling one being Optional the other being Required etc).
+2. `ModelToModel` - which specifies that this Schema Field should be mapped onto another Model. This happens when the Fields `ObjectDefinition` is either a `Reference`, `Dictionary[string, Reference]`, `List[Reference]` or `Set[Reference]`.
+3. `Manual` - specifies that this Field must be manually mapped (for example due to complex behaviour).
+
+> _Future thought:_ additional Field Mapping types could be added to support additional use-cases. For example, transforming values (e.g. such as having an `Integer` used in the Terraform Schema, but a `String` exposed in the SDK - or to support handling `SkuName` etc). 
+
+### a. `DirectAssignment` Field Mappings
+
+A `DirectAssignment` mapping specifies that the value of a Field within the Terraform Schema should be mapped onto the value of a Field within the related (Go) SDK Model - and vica-versa.
```
stephybun, line 33:
```suggestion
A `DirectAssignment` mapping specifies that the value of a Field within the Terraform Schema should be mapped onto the value of a Field within the related (Go) SDK Model and vice-versa.
```

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+    }
+  ]
+}
+```
+
+### b. `ModelToModel` Field Mappings
+
+ModelToModel Field Mappings - not to be confused with ModelToModel Mappings (below) - allow mapping a Terraform Schema Model onto a field within an SDK Model.
+
+The best example of this is allowing the top-level Terraform Schema Model for a given Terraform Resource to map onto the `properties` object that exists in most payloads in Azure Resource Manager - allowing for fields which exist at the root of the Terraform Schema to be mapped to a nested object.
```
stephybun, line 74: I think it would be helpful to give explicit examples here for what top-level Terraform Schema Models you mean, e.g. `location`? `tags`?

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+* `schemaModelName` - specifies the name of the Terraform Schema Model where this value should be mapped from.
+* `schemaFieldPath` - specifies the path to the Field within the Terraform Schema Model defined in `schemaModelName` which this should be mapped from.
+* `sdkModelName` - specifies the name of the SDK Model where this value should be mapped onto.
+* `sdkFieldPath` - specifies the Path to the Field within the SdkModel where the Schema Field should be mapped onto.
+
+> **Note:** Whilst `schemaFieldPath` and `sdkFieldPath` are named `Path` to allow supporting indexed keys in the future (for example using `AppSettings.SomeSetting` to retrieve the value for the key `SomeSetting` within the Dictionary defined in the `AppSettings` property) - at this point in time these only support top-level fields (e.g. `Name`).
+
+An example of a Direct Assignment Mapping can be seen below:
+
+> **Note:** A more complete example can be seen under `Examples` below.
```
stephybun, line 52: I understand this is trying to be helpful but repeating it multiple times throughout this guide is unnecessary bloat and noise for a topic that's complex. Say it once, probably at the top of the document.

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+ModelToModel Field Mappings - not to be confused with ModelToModel Mappings (below) - allow mapping a Terraform Schema Model onto a field within an SDK Model.
+
+The best example of this is allowing the top-level Terraform Schema Model for a given Terraform Resource to map onto the `properties` object that exists in most payloads in Azure Resource Manager - allowing for fields which exist at the root of the Terraform Schema to be mapped to a nested object.
+
+Each ModelToModel Field Mapping must have an associated ModelToModel mapping, these are used together:
+
+* A ModelToModel Field Mapping is used to output the call to the mapping function itself.
+* A ModelToModel Mapping is used to output the mapping function, the fields which should be mapped are then inferred from the ModelToModel Field Mappings.
+
+> **Note:** A complete end-to-end example can be found below under `Example`.
```
stephybun, line 81: I understand this is trying to be helpful but repeating it multiple times throughout this guide is unnecessary bloat and noise for a topic that's complex. Say it once, probably at the top of the document.

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+  "modelToModelMappings": [
+    {
+      "schemaModelName": "VirtualMachineResourceSchema",
+      "sdkModelName": "VirtualMachine"
+    }
+  ]
+}
+```
+
+> **Note:** An end-to-end example can be found under `Example` below.
```
stephybun, line 190: I understand this is trying to be helpful but repeating it multiple times throughout this guide is unnecessary bloat and noise for a topic that's complex. Say it once, probably at the top of the document.

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+### b. `ModelToModel` Field Mappings
+
+ModelToModel Field Mappings - not to be confused with ModelToModel Mappings (below) - allow mapping a Terraform Schema Model onto a field within an SDK Model.
+
+The best example of this is allowing the top-level Terraform Schema Model for a given Terraform Resource to map onto the `properties` object that exists in most payloads in Azure Resource Manager - allowing for fields which exist at the root of the Terraform Schema to be mapped to a nested object.
+
+Each ModelToModel Field Mapping must have an associated ModelToModel mapping, these are used together:
+
+* A ModelToModel Field Mapping is used to output the call to the mapping function itself.
+* A ModelToModel Mapping is used to output the mapping function, the fields which should be mapped are then inferred from the ModelToModel Field Mappings.
```
stephybun, line 79: This sounds like they're both doing the same thing: 
`output the call to the mapping function` vs. `output the mapping function`
This might want to be rephrased or more information to distinguish the difference in functionality.

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+
+> **Note:** A complete end-to-end example can be found below under `Example`.
+
+Model to Model Field Mappings support the following properties:
```
stephybun, line 83: We're not separating this anywhere else in the docs, to be consistent I think this should be 
```suggestion
ModelToModel Field Mappings support the following properties:
```

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+
+* `schemaModelName` - specifies the name of the Terraform Schema Model that should be mapped onto the Field within the SDK Model.
+* `sdkModelName` - specifies the name of the SDK Model containing the SDK Field that the Terraform Schema Model should be mapped onto.
+* `sdkFieldPath` - specifies the Path to the Field within the SDK Model where the Terraform Schema Model should be mapped onto.
+
+An example of a Model to Model Field Mapping can be seen below:
```
stephybun, line 89: Same here 
```suggestion
An example of a ModelToModel Field Mapping can be seen below: 
```

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+func (r VirtualMachineResource) mapVirtualMachineResourceSchemaToVirtualMachineProperties(input VirtualMachineResourceSchema, output *virtualmachines.VirtualMachine) error {
+    if output.Properties == nil {
+        output.Properties = &virtualmachines.VirtualMachineProperties{}
+    }
+    if err := r.mapVirtualMachineResourceSchemaToVirtualMachineProperties(input, output.Properties); err != nil {
+        return fmt.Errorf("mapping Schema to SDK Field %q / Model %q: %+v", "Properties", "SolarSystemProperties", err)
+    }
+    
+    return nil
+}
```
stephybun, line 118: In this example this function would be calling itself, is that what it's supposed to do?

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+As mentioned above, this implies the presence of an associated ModelToModel Mapping, which would then generate the mapping for the Model itself:
+
+```go
+func (r VirtualMachineResource) mapVirtualMachineResourceSchemaToVirtualMachineProperties(input VirtualMachineResourceSchema, output *virtualmachines.VirtualMachineProperties) error {
+    // ...
+    // DirectAssignment Field Mappings would be output here
+    // ...
+    
+    return nil
+}
```
stephybun, line 130: From the explanation of the example this would be generated in addition to the function example above, but it has the same function name as the function above?

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+```
+
+### c. `Manual` Field Mappings
+
+> **Note:** At this point in time Manual Mappings are only partially configured and **need reworking** - as such this section explains the intention this type of Mapping, so that we can fix this in the future.
```
stephybun, line 135:
```suggestion
> **Note:** At this point in time Manual Mappings are only partially configured and **need reworking** - as such this section explains the intention of this type of Mapping, so that we can fix this in the future.
```

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+In this example the following properties would exist:
+
+* `schemaModelName` - specifies the name of the Terraform Schema Model where the Field that should be mapped to/from exists.
+* `schemaFieldPath` - specifies the path to the Field within the Terraform Schema Model that should be mapped to/from the Field within the SDK.
+* `schemaToSdkMethodName` - specifies the name of a function that should be assumed to exist within the Terraform Provider, which maps the value of the Terraform Schema Field onto the SDK Field.
+* `sdkModelName` - specifies the name of the SDK Model where the SDK Field defined in `sdkFieldPath` exists.
+* `sdkFieldPath`- specifies the path to the Field within the SDK Model where the Schema Field should be mapped to/from.
+* `sdkToSchemaMethodName` - specifies the name of a function that should be assumed to exist within the Terraform Provider, which maps the value of the SDK Field onto the Terraform Schema Field.
+
+> **Note:** The current implementation that exists exposed only `methodName` - as such is only useful for one-way value transformations.
```
stephybun, line 164:
```suggestion
> **Note:** The current implementation that exists exposes only `methodName` and is only useful for one-way value transformations.
```

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+
+## 2: Model to Model Mappings
```
stephybun, line 166: Inconsistent with the other headers in this doc
```suggestion
## 2: `ModelToModel` Mappings
```

docs/internals-mappings.md
```diff
@@ -0,0 +1,428 @@
...
+* `fieldMappings[2]` - The Terraform Schema Field `planets` (represented by the field `Planets` within the Schema Model `SolarSystemResourceSchema`) gets mapped onto the SDK Field `Planets` within the SDK Model `SolarSystemProperties`. Since this is a list, each item will be mapped across in-turn.``
+* `fieldMappings[3]` - The Terraform Schema Field `name` within `planets` (represented by the field `Name` within the Schema Model `SolarSystemResourcePlanetSchema`) gets mapped onto the SDK Field `Name` within the SDK Model `Planet`.
+
+> **Note:** In this (simplified) example there are no `resourceIdMappings` - however examples of this can be seen above.
+
+This would allow the Terraform Generator to output mapping functions similar to below:
+
+> Whilst mappings are bidirectional (that is Schema Model -> SDK Model and SDK Model -> Schema Model) - to reduce the amount of output only the Schema Model -> SDK Model mappings are shown below - but the inverse can be implied.
```
stephybun, line 391: Leaving things to be implied for complex subjects will lead to misinterpretations, misunderstandings and errors as we've learnt. For completeness the SDK Model -> Schema Model mappings should be included.

stephybun (changes requested):
Thanks @tombuildsstuff for the detailed explanation and examples.

I left some suggestions which would be good to have addressed before we merge this.

tombuildsstuff:
In the end we ended up deciding to document and rename these inline within the new SDK types, so I'm going to close this out:

* [Terraform Field Mapping Definition](https://github.com/hashicorp/pandora/blob/main/tools/data-api-sdk/v1/models/terraform_field_mapping_definition.go)
* [Terraform Model to Model Field Mapping Definition](https://github.com/hashicorp/pandora/blob/main/tools/data-api-sdk/v1/models/terraform_model_to_model_field_mapping_definition.go)
* [Terraform Resource ID Mapping Definition](https://github.com/hashicorp/pandora/blob/main/tools/data-api-sdk/v1/models/terraform_resource_id_mapping_definition.go)

stephybun:
@tombuildsstuff I believe what's written here is valuable and helpful for understanding mappings and the detailed explanations and examples in it are not substituted by the comments left in the code that are linked in your closing comment for this PR. 

I've restored this branch since I don't believe this information should be deleted entirely and would like this to be re-opened, without the expectation that you are required to finish this off.

PR #3268, opened 2023-11-01
DataAPI V2: Map Terraform information

mbfrahry:
Supersedes #3214 

For some reason `mappings` isn't coming through to the api though the information is there. That'll be one to hunt down and there are a few todos to figure out.


stephybun (commented):
Thanks for this PR @mbfrahry - I made a first pass through and its looking good, I left a few minor comments which would be good to fix up.

I think this is in a state to merge and begin comparing the data to the former data API to iron out the bugs

katbyte (approved):
LGTM 🍬

PR #3274, opened 2023-11-02
DataAPIV2: Add additional error messaging when loading files and add additional ObjectDefinitionTypes 

mbfrahry:
Data API V2 is currently failing on main. 

I've added additional error messaging to see where we were going wrong and added additional ObjectDefinitionTypes to fix what was broken


katbyte (approved):
LGTM 🌵

PR #3291, opened 2023-11-06
dataapiv2 - Various Fixes to the terraform side

mbfrahry:
This PR fixes some of the differences between the Terraform side of the data api v1 vs v2

### DIFF
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
@@ -20,14 +20,11 @@ func buildTerraformResourceDefinition(resourceLabel string, input resourcemanage
 			Name:             input.DeleteMethod.MethodName,
 			TimeoutInMinutes: input.DeleteMethod.TimeoutInMinutes,
 		},
-		Description:                  input.Documentation.Description,
-		DisplayName:                  input.DisplayName,
-		ExampleUsage:                 input.Documentation.ExampleUsageHcl,
-		Generate:                     input.Generate,
-		GenerateIdValidationFunction: input.GenerateIdValidation,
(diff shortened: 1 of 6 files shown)


tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
```diff
@@ -20,21 +20,23 @@ func buildTerraformResourceDefinition(resourceLabel string, input resourcemanage
...
+		Label:        resourceLabel,
 		ReadMethod: dataApiModels.TerraformMethodDefinition{
 			Generate:         input.ReadMethod.Generate,
 			Name:             input.ReadMethod.MethodName,
 			TimeoutInMinutes: input.ReadMethod.TimeoutInMinutes,
 		},
 		ResourceIdName: input.ResourceIdName,
 		UpdateMethod:   nil,
+
+		// todo these are hardcoded to true in the C# generation. Does that hold true here?
```
tombuildsstuff, line 36: Yep, these aren't exposed right now but are cheap placeholders for when/if we need those in the future

However these should be hard-coded up one level, where the source data is passed in, so we can update these to use the existing `input.GenerateIdValidation` (etc) values and set this up one level if necessary?

tombuildsstuff (commented):
One comment to ensure these are threaded through consistently, but this otherwise LGTM 👍

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
mbfrahry: Done!

stephybun (approved):
Thanks @mbfrahry LGTM 👍

tombuildsstuff (approved):
LGTM 👍

PR #3298, opened 2023-11-07
Data API V2: thread through `TypeHintIn`, `Description` and add validation around discriminators

stephybun:
When loading and parsing the data we now just map everything from the data definitions over into the data api models 1:1, then do additional validation around discriminators when we validate object definitions etc.

This will also include various fixes to the data that's output in the data API. See individual commits.

### DIFF
--- a/tools/data-api/internal/repositories/services.go
+++ b/tools/data-api/internal/repositories/services.go
@@ -346,8 +346,6 @@ func parseModelFromFilePath(filePath string) (*ModelDetails, error) {
 			Optional:    field.Optional,
 			Required:    field.Required,
 			Description: field.Description,
-			// TODO this field is still missing - where is this information taken from?
-			//ForceNew:         false,
 		}
 
 		objectDefinition, err := mapObjectDefinition(&field.ObjectDefinition)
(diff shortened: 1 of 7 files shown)


tools/data-api/internal/repositories/services.go
```diff
@@ -337,18 +338,15 @@ func parseModelFromFilePath(filePath string) (*ModelDetails, error) {
...
-			//Default:          nil,
-			// TODO exposed when other #3238 is merged
-			//Description:      "",
+			IsTypeHint:  field.ContainsDiscriminatedTypeValue,
+			JsonName:    field.JsonName,
+			Optional:    field.Optional,
+			Required:    field.Required,
+			Description: field.Description,
+			// TODO this field is still missing - where is this information taken from?
 			//ForceNew:         false,
```
tombuildsstuff, line 350: This isn't actually used today - so we should be able to remove this one.

This stems from a period where we were investigating pushing the Terraform details into the SDK models, but that design changed and this wasn't cleaned up - so this remnant can be safely removed

tombuildsstuff (approved):
A couple of minor comments but otherwise 👍

tombuildsstuff (approved):
LGTM 👍

PR #3299, opened 2023-11-07
dataapiv2 - fix issue where partial data is loaded

### LINKED ISSUE #3293: Data API V2: Services that fail to load only error the first time
When loading the Services fails in Data API V2, the first request will return a 500 (since the data isn't valid) - however subsequent requests will serve a partial data-set:

`GET http://localhost:8080/v1/resource-manager/services`

```json
{
  "services": {
    "AAD": {
      "generate": false,
      "uri": "/v1/resource-manager/services/AAD"
    },
    "Advisor": {
      "generate": false,
      "uri": "/v1/resource-manager/services/Advisor"
    }
  }
}
```

We should fix this so that we're not setting the data until it's fully loaded to avoid this.

mbfrahry:
Fixes #3293

### DIFF
--- a/tools/data-api/internal/repositories/services.go
+++ b/tools/data-api/internal/repositories/services.go
@@ -80,6 +80,7 @@ func (s *ServicesRepositoryImpl) GetAll(serviceType ServiceType) (*[]ServiceDeta
 		}
 
 		if services != nil {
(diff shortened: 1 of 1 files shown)


stephybun (approved):
Thanks @mbfrahry LGTM 🚁

PR #3309, opened 2023-11-09
Config: add ScVmm@2023-10-07

neil-yechenwei:
We intend to support ScVmm(alias: System Center Virtual Machine Manager) RP. This is the API(https://github.com/Azure/azure-rest-api-specs/blob/main/specification/scvmm/resource-manager/Microsoft.ScVmm/stable/2023-10-07/scvmm.json).


config/resource-manager.hcl
```diff
@@ -468,6 +468,10 @@ service "resources" {
   name      = "Resources"
   available = ["2020-05-01", "2020-06-01", "2020-10-01", "2021-04-01", "2021-06-01", "2021-07-01", "2022-02-01", "2022-06-01", "2022-07-01-preview", "2022-09-01", "2022-12-01", "2023-04-01", "2023-07-01", "2023-08-01"]
 }
+service "scvmm" {
+  name      = "ScVmm"
```
tombuildsstuff, line 472: this should be:

```suggestion
  name      = "SystemCenterVirtualMachineManager"
```

tombuildsstuff (changes requested):
hi @neil-yechenwei 

Rather than using the name `ScVmm` we should make this `SystemCenterVirtualMachineManager` - but if we can fix that up this otherwise LGTM 👍

config/resource-manager.hcl
neil-yechenwei: Thanks for the suggestion. I updated it.

tombuildsstuff (approved):

PR #3312, opened 2023-11-09
Data API V2: set fields correctly back into model and various fixes/cleanup

stephybun:
The mapped fields from the data definitions weren't being set correctly.

Also exposing the `ParentTypeName` in the v1 endpoint and removed `Default` in the v1 endpoint since this is not used.

### DIFF
--- a/tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
+++ b/tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
@@ -17,7 +17,7 @@ func codeForOperation(operationName string, input importerModels.OperationDetail
 	sort.Ints(expectedStatusCodes)
 	output := dataApiModels.Operation{
 		Name:                             operationName,
-		ContentType:                      input.ContentType,
+		ContentType:                      fmt.Sprintf("%s; charset=utf-8", input.ContentType),
 		ExpectedStatusCodes:              expectedStatusCodes,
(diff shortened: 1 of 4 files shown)


tombuildsstuff (approved):

tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go
```diff
@@ -17,7 +17,7 @@ func codeForOperation(operationName string, input importerModels.OperationDetail
 	sort.Ints(expectedStatusCodes)
 	output := dataApiModels.Operation{
 		Name:                             operationName,
-		ContentType:                      input.ContentType,
+		ContentType:                      fmt.Sprintf("%s; charset=utf-8", input.ContentType),
```
tombuildsstuff, line 20: FWIW this'll be a breaking change in the case of APIs which don't use JSON ([Automation](https://github.com/hashicorp/pandora/blob/104b9f61066942b6760002875fcfd382b88ab960/data/Pandora.Definitions.ResourceManager/Automation/v2023_11_01/RunbookDraft/Operation-ReplaceContent.cs#L17) being an example that uses a non-JSON payload) - I don't think that's _necessarily_ a bad thing - but we should validate that doesn't break things.

Which makes me wonder if we should be making this change in the importer instead, so that we can roll that out with the V1 API?

PR #3321, opened 2023-11-10
config: removing older versions of `redisenterprise`

tombuildsstuff:
These are no longer needed as of https://github.com/hashicorp/terraform-provider-azurerm/pull/23865


tombuildsstuff:
The newer API version still isn't live - closing this for now.

Issue #3315, opened 2023-11-09
Data API V2: Custom differ


stephybun:
The auto generated PRs that regenerate the Data API definitions often contain many changes that are time consuming to review and validate. We need an automated way to summarise the changes made and presented in a user friendly manner to expedite the review process of these changes and to reduce the chance of errors or problems being overlooked and creeping their way into the provider.

Common scenarios that need to be covered by this differ are:
- Extracting a unique list of resource IDs that have been changed/added so that the segments can be checked for the correct casing (this functionality can be taken from the existing [`extract-tf-resource-ids` job](https://github.com/hashicorp/pandora/tree/main/tools/extract-tf-resource-ids))
- When definitions are removed in a service's API version, automatically detecting whether the provider imports and uses this version and any of these endpoints that have been modified/removed and to flag breaking changes

tombuildsstuff:
Calling out a design limitation from the current setup / a topic that we've previously talked about here too:

The V1 tool (`extract-tf-resource-ids`) currently has a hard-dependency on the GitHub API which means this tool can't be used locally.

In addition the GitHub API uses pagination when the changeset associated with a pull request contains over 3000 files changes, which the V1 tool isn't accounting for (meaning that we're not outputting all the changes today, [example](https://github.com/hashicorp/pandora/pull/3322)) - and whilst we can update the tool to handle pagination, switching to a local setup removes this limitation.

As such rather than relying on the contents of local files, presumably it'd be better to query the Data API to get the data from both `main` and the current branch to allow us to perform the diff, which presumably would be easiest by having the tool doing something along these lines:

1. Clone a copy of the `main` branch from the current working directory (e.g. this repository - intentionally not `main` from GitHub) into a temporary directory. (e.g. `git clone -b main /path/to/working/directory /temp/directory`, but presumably [in code](https://github.com/go-git/go-git/blob/a9509b0a52f116ea2969d51890a4265d98d50dcb/_examples/clone/main.go#L20-L23) which [looks to be possible via the options](https://github.com/go-git/go-git/blob/a9509b0a52f116ea2969d51890a4265d98d50dcb/options.go#L48)).
2. Spinning up an instance of the Data API from that branch (to give us a baseline) - specifying the port the Data API should be using as an Environment Variable.
3. Having the tool retrieve the changes from the `main` version of the Data API.
4. Shutting that version of the Data API down.
5. Launching the Data API from the current working directory - specifying the port the Data API should be using as an Environment Variable.
[...]

PR #3340, opened 2023-11-13
`tools/extract-tf-resource-ids`: support for large changesets

tombuildsstuff:
This PR updates the `extract-tf-resource-id` tool such that it works with larger sets of changes which is necessary since both sets of API Definitions are currently being output. This fixes an issue seen in #3339 where changesets containing over 3000 files changed wouldn't get a diff output.

Since this'll be replaced in the not-too-distant future by https://github.com/hashicorp/pandora/issues/3315 - I've opted just to pull the raw diff (which is complete) rather than iterating over the pages (which doesn't appear to be supported by the GitHub SDK)

This PR also updated the GitHub Action, since it appears the file trigger/filter only takes note of the first 3000 files changed - and not the entire changeset - which should ensure this triggers for any API Definition regenerations until #3315 is implemented

Example comment using this PR: https://github.com/hashicorp/pandora/pull/3339#issuecomment-1808646339


mbfrahry (approved):
LGTM

PR #3341, opened 2023-11-14
importer-rest-api-specs - Output Terraform Tests as HCL

mbfrahry:
This PR swaps how we output Terraform tests by outputting those tests as HCL rather than json


stephybun (approved):
Thanks @mbfrahry LGTM 💯

PR #3343, opened 2023-11-14
automation: running the automation when `./api-definitions` changes

tombuildsstuff:
This works around a limitation where GHA only triggers based on the first 3000 files in a given changeset, which may or may not contain `./data`. Since API Definitions are temporarily output to both `./api-definitions` and `./data` - this PR means we trigger on both regardless, working around this issue for the moment.


jackofallops (approved):
LGTM 👍

PR #3353, opened 2023-11-15
Data API V2: Make adjustments to align data between v1 and v2

stephybun:


tombuildsstuff (approved):
As discussed offline, there's a couple of things we should look to fix out-of-band, but otherwise this LGTM 👍

PR #3368, opened 2023-11-16
wrapper-automation - Add args.UseV2Generator

mbfrahry:
This PR adds a flag to `wrapper-automation` and the scripts that call `wrapper-automation` that will flips between the old and new data api


stephybun (approved):
Thanks @mbfrahry LGTM 🐢

PR #3375, opened 2023-11-17
Data API V2: Refactor the data API models under `sdk/dataapimodels`

stephybun:
I chose to put these under the folder `dataapimodels` instead of `v1models` since I think the term v1 is unclear and has the potential to be confusing given that we use v1 for versioning several things, e.g. endpoint within the data api as well as the old and new versions of the data api (data api v2)

If we need to add more models for a new version of the endpoint e.g. v2 we can reconsider the naming then. But for the forseeable future this is sufficient and clearer.

The removal of the models in the `data-api` tool will happen in a follow up PR when these models can be vendored into the project.


jackofallops (approved):
LGTM 👍

Issue #3384, opened 2023-11-20
`Automation/2023-11-01` lacks `Delete` operation for `hybridrunbookworkergroup`


aristosvo:
While trying to enable [PowerShell 7.2](https://github.com/hashicorp/terraform-provider-azurerm/issues/14089) for the `automation_runbook` and `automation_module` by moving from `Automation/2022-08-08` to `Automation/2023-11-01`, I found out that the `Delete` operation for the `hybridrunbookworkergroup` client was not generated:

```
internal/services/automation/automation_hybrid_runbook_worker_group.go:184:23: client.Delete undefined (type *hybridrunbookworkergroup.HybridRunbookWorkerGroupClient has no field or method Delete)
```

It is not defined in the Data API generated by Pandora, although it is defined in the REST API specs (https://github.com/sushil490023/azure-rest-api-specs/blob/0dd49a444195fef7f3555cad038cb7665cbd928c/specification/automation/resource-manager/Microsoft.Automation/stable/2023-11-01/hybridRunbookWorkerGroup.json#L40)

**Edit1**:

The issue is probably the tag in de `delete` operation, which contains `ListAllHybridRunbookWorkerGroupInAutomationAccount` instead of `HybridRunbookWorkerGroup`.

Don't know how to fix that properly though, but at least the issue is in the schema 😟

**Edit2**:
It is a recurring issue:

- https://github.com/Azure/azure-rest-api-specs/pull/25108

At the same time another issue is still in place:

- https://github.com/Azure/azure-rest-api-specs/pull/25435

aristosvo:
The issue is probably the tag in de `delete` operation, which contains `ListAllHybridRunbookWorkerGroupInAutomationAccount` instead of `HybridRunbookWorkerGroup`.

Don't know how to fix that properly though, but at least the issue is in the schema 😟

aristosvo:
It is a recurring issue: 
- https://github.com/Azure/azure-rest-api-specs/pull/25108

At  the same time another issue is still in place
- https://github.com/Azure/azure-rest-api-specs/pull/25435

aristosvo:
Fixed by #3385

PR #3385, opened 2023-11-20
workarounds(`Automation`): recurrence of old issues with the spec in new 2023-version

aristosvo:
Fixes issue #3384

I've submitted new files (and filed new PRs for this version), please let me know if we prefer to merge this to single files per problem. (LRO python3 in one file for both versions, tag on delete operation in one file for both versions)


tombuildsstuff (approved):
LGTM 👍

tombuildsstuff (changes requested):
@aristosvo thanks for this PR.

Chatting internally about this PR, rather than duplicating the Data Workaround/upstream PR to fix this issue in the new API version, we'd prefer to have a single Data Workaround and Upstream PR per issue, since we're tracking those on our side/running them down with MSFT.

We should be able to [reuse this existing approach to apply this Data Workaround to multiple versions of the API](https://github.com/hashicorp/pandora/blob/f8c422b3d57b0112f830deea504e3b5c14f5e444/tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_batch_21291.go#L18-L20) - and in order to support that I hope you don't mind but I'm going to rebase/update the two existing upstream PRs to include your commits, meaning that we've got a single PR tracking that issue across multiple versions of the API (which means we can chase a single PR with MSFT/the Service Team).

As such, would you mind updating this PR so that we're applying the existing Data Workaround to both API Versions instead? Once that's done I think this should otherwise be good to go, however 👍

aristosvo:
Hi @tombuildsstuff !

Thanks for the remark! I was kinda expecting that, as that makes tracking from your side easier. I will combine the files tonight.

tombuildsstuff (approved):

PR #3395, opened 2023-11-22
tools/data-api: using a go mod replace to find the sdk

tombuildsstuff:
This ensures that these are sourced from the local disk rather than an untagged revision - to match the other tooling.

ctx: https://github.com/hashicorp/pandora/pull/3382/files#diff-cc49cf4a51c8b072add49b7bf868102dfa5d7601027382d31522f4bdf33ca59aR10


jackofallops (approved):
LGTM 👍

PR #3401, opened 2023-11-24
Config: add ExtendedLocation@2021-08-15

neil-yechenwei:
As SCVMM.VMMServer(System Center Virtual Machine Manager Server) resource depends on [ExtendedLocation.CustomLocation](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/extendedlocation/resource-manager/Microsoft.ExtendedLocation/stable/2021-08-15/customlocations.json), so it needs to be added.


neil-yechenwei:
I will close this PR since service team confirmed that testing System Center Virtual Machine Manager Server requires additional complicated configurations on the created customLocation resource so that we have to pass the required values of the existing customLocation while running test case about SCVMM Server.

PR #3416, opened 2023-11-29
New Resource: generating Kubernetes Fleet Member

ms-henglu:


zioproto:
Related to https://github.com/hashicorp/terraform-provider-azurerm/issues/21468

stephybun (changes requested):
Thanks for trying to generate this resource with Pandora @ms-henglu, it is a good candidate for that.

In addition to the comments left in-line the following also needs to be done:
1. Rebase this PR - a lot of new functionality, changes and fixes have been made since this was opened, when we try and generate this resource with the current state of the repo the generation is skipped (this was a known bug that was fixed)
2. We need to update the [common ID for fleet in `go-azure-helpers`](https://github.com/hashicorp/go-azure-helpers/blob/main/resourcemanager/commonids/fleet.go) to `KubernetesFleetId`, this is related to the comment left in-line on the `fleet_id` test dependency
3. We need to add a `common_id_kubernetes_fleet_id.go` file to [this directory](https://github.com/hashicorp/pandora/tree/main/tools/importer-rest-api-specs/components/parser/resourceids) and also add this new common ID to the `common_ids.go` file

ms-henglu:
Hi @stephybun ,

About the review comment "changing fleet_id to kubernetes_fleet_id": https://github.com/hashicorp/pandora/pull/3416#discussion_r1472498447

I did some tests and found that `fleet_id` is the generated field name and I can't override it to `kubernetes_fleet_id `  by
```hcl
        overrides "fleet_id" {
          updated_name = "kubernetes_fleet_id"
        }
```

I also checked the codes, it seems resource id fields doesn't support the `overrides` feature to update the name: https://github.com/hashicorp/pandora/blob/main/tools/importer-rest-api-specs/components/schema/fields_resource_id.go#L59

Please correct me if I was missing something. Thanks!

ms-henglu:
Thanks @stephybun  for the quick fix! I've rebased the PR, please take another look.

stephybun:
I'd be happy to, but not all of the review points have been addressed, mainly point 2. and 3. in [this comment](https://github.com/hashicorp/pandora/pull/3416#pullrequestreview-1853173912) at the very least 3. needs to be done otherwise the common ID won't be applied.

ms-henglu:
To address the point 2: https://github.com/hashicorp/go-azure-helpers/pull/212

PR #3423, opened 2023-12-01
dataapiv2: Various terraform fixes

mbfrahry:


tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_terraform_resource_definition.go
```diff
@@ -22,7 +24,7 @@ func buildTerraformResourceDefinition(resourceLabel string, input resourcemanage
 		},
 		Description:                  input.Documentation.Description,
 		DisplayName:                  input.DisplayName,
-		ExampleUsage:                 input.Documentation.ExampleUsageHcl,
+		ExampleUsage:                 strings.TrimPrefix(strings.TrimSuffix(input.Documentation.ExampleUsageHcl, "\n"), "\n"),
```
stephybun, line 27: I suspect, we might want to manipulate the formatting of this in the parser instead of in here. But for now I think this is fine, maybe add a `// todo` above it?

stephybun (approved):
Thanks @mbfrahry LGTM 👍

tombuildsstuff (commented):
Spotted one thing whilst passing through but otherwise 👍

PR #3426, opened 2023-12-01
Data API V2: fixing various data issues in the parser and exposing missing attributes in the service and version definitions

### LINKED ISSUE #3363: Data API V2: Some operations are missing status codes
In the Data API we've had to add logic to add default status codes to operations (to prevent a diff in the generated sdk with the v2 data) because some services e.g. `SecurityInsights` have operations where this information is missing.

This is potentially a swagger bug/should be handled in the swagger parser of the importer.

Will cross link more information once #3353 is merged.

### LINKED ISSUE #3364: Data API V2: Outputting the response object for all operations
Currently in the `dataapigeneratorjson` we discard the response object for certain types of operations, e.g. long running ones or if it's not a list operation

https://github.com/hashicorp/pandora/blob/60439bc2419ea4b745579a176a777b664b2cbcd9/tools/importer-rest-api-specs/components/dataapigeneratorjson/templater_operations.go#L42-L53

This is done because it is also the logic that exists in the current Data API running on C#

https://github.com/hashicorp/pandora/blob/60439bc2419ea4b745579a176a777b664b2cbcd9/data/Pandora.Data/Transformers/Operation.cs#L39-L56

TODO: 
1.  A large amount of changes are required to support returning the response object for LRO's - we will not do this, instead we should strip the response object for LRO's in the swagger parser
2. We should return the response object for all other operations, once 1. is done we just need to map responseObject over as is.

stephybun:
This PR fixes the following:

- fixes #3364
- fixes #3363
- maps the value for generate over in the service details (for the `generate-go-sdk` tool)
- loads the `ApiVersionDefinition.json` in `data-api` so that we can expose the source information (also for the `generate-go-sdk` tool)

Everything has been tested and generated locally:
- the `go-azure-sdk` generates successfully with these changes and without any diffs
- this PR will trigger a regen of the api-definitions, but the only changes should be the addition of the `Source` field to all of the `ApiVersionDefinitions.json` in the `api-definitions` folder

### DIFF
--- a/tools/data-api/internal/repositories/mappings.go
+++ b/tools/data-api/internal/repositories/mappings.go
@@ -153,3 +153,16 @@ func mapResourceIdSegmentType(input dataapimodels.ResourceIdSegmentType) (*Resou
 
 	return nil, fmt.Errorf("unmapped Resource Id Segment Type %q", string(input))
 }
+
+func mapApiDefinitionSourceType(input dataapimodels.ApiDefinitionsSource) (*ApiDefinitionSourceType, error) {
+	mappings := map[dataapimodels.ApiDefinitionsSource]ApiDefinitionSourceType{
+		dataapimodels.HandWrittenApiDefinitionsSource:                      HandWrittenApiDefinitionsSource,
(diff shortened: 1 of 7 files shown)


mbfrahry (approved):
LGTM

PR #3428, opened 2023-12-01
dataapiv2 - Add github actions to dataapiv2

### LINKED ISSUE #3366: Data API v2: Validation Command/Unit Test
When Github Actions run, to be able to validate that the API Definitions are correct we currently run a unit test via `make test` to validate these parse correctly.

At this point in time whilst Data API v2 does this when loading the repository, this isn't available via a unit test/command - so it'd be worth adding:

1. A unit test to do this
2. A command do this (for the CLI)?

mbfrahry:
Fixes #3366

### DIFF
--- a/tools/data-api/internal/repositories/services_test.go
+++ b/tools/data-api/internal/repositories/services_test.go
@@ -15,7 +15,7 @@ func TestServices_ResourceManager(t *testing.T) {
 	}
 }
 
-func TestServices_Graph(t *testing.T) {
+func TestServices_MicrosoftGraph(t *testing.T) {
 	repo, err := NewServicesRepository("../../../../api-definitions/", MicrosoftGraphV1StableServiceType, nil)
 	if err != nil {
 		t.Fatalf(err.Error())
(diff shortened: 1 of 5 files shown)


stephybun (approved):
Thanks @mbfrahry LGTM 👍

tools/data-api/internal/repositories/services_test.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+
+	if _, err := repo.GetAll(ResourceManagerServiceType); err != nil {
+		t.Fatalf(err.Error())
+	}
+}
+
+func TestServices_Graph(t *testing.T) {
```
tombuildsstuff, line 18: To be pedantic, can we keep this consistent as `MicrosoftGraph`? There's different kinds of `Graph` referenced in Azure - (AAD Graph, MS Graph, Resource Graph) so since we're being consistent elsewhere it'd be nice to be consistent here too:

```suggestion
func TestServices_MicrosoftGraph(t *testing.T) {
```

tombuildsstuff (approved):
One super minor thing around naming consistency, but this otherwise LGTM 👍

tools/data-api/internal/repositories/services_test.go
mbfrahry: Good shout! Ty ty

PR #3462, opened 2023-12-07
New Tool: Data API Differ

### LINKED ISSUE #3315: Data API V2: Custom differ

tombuildsstuff:
This PR introduces a new tool `data-api-differ` which detects and then outputs the changes between two different sets of API Definitions.

The intention here is to provide a replacement for the older `extract-resource-id-segments` tool - but also to output a more granular diff of what’s happened so that anyone looking at a PR which updates the API Definitions has a clear understanding of what’s changing in that Pull Request.

This tool detects three types of changes:

1. **Breaking Changes** - for example when a Service/API Version/API Resource/Constant/Field/Model/Operation/Resource ID gets removed, or when a Resource ID/Options object gets added to an Operation.
2. **Breaking and Non-Breaking Changes** - detects and outputs both the Breaking and Non-Breaking Changes - for example where an API Version gets added, a new Constant Key/Value gets added or a Service gets removed.
3. **New Static Identifiers present within New Resource IDs** - this determines any Static Identifiers (e.g. the name of the resource provider or the fixed value for a Static Resource ID Segment) and then outputs a unique, sorted list.

A shortened example of the output for each of these commands can be seen in the README - but [the output from an earlier build can be found in this Gist](https://gist.github.com/tombuildsstuff/94c934df0f3555c68627da9c07565490) for review.

This tool detects each Type of Change as unique struct - meaning that it’s possible to quickly add more granular reports as needed - and I suspect we’ll want to extend this to support diffing two different Versions of two APIs (e.g. `2020-01-01` and `2023-06-01`) in the future.

This PR also adds the associated automation, running this tool against Pull Requests to be able to detect changes between `main` and a given Pull Request.

Fixes #3315

### Notes / Limitations

* When adding a new Service/API Version at this point in time the _entire_ diff is output in one long block - it’s likely we can change how this is presented to make this easier to review.
* At this point in time the tool is limited to Resource Manager until the Data API SDK is updated and consolidated - but that should be easy to extend in the future (once Data API V2 is fully rolled out). Presumably we’ll want to group the Changes against by Data Source (Microsoft Graph / Resource Manager) when that happens.
* At this point the changes being output assume an understanding of the values (e.g. `Model has a new Parent Type`) has no backing documentation. We should look to add documentation to `./docs` which explains what each of the types of changes are, and the consequence of this/if these are resolved.
* At this point time this isn’t checking if the Services are used within `hashicorp/terraform-provider-azurerm` (or `hashicorp/terraform-provider-azuread`) but that seems like a logical extension to be able to determine when a breaking change is _fine_ (e.g. removing an unused old Service/API Version) and when it’s problematic (e.g. removing a Service/API Version which is currently vendored into the Provider[s]).
* At this point in time this tool doesn’t detect changes to the Terraform Resources, that’s left as a logical extension point for the future.
* Arguably the tool could do with an updated name, @stephybun suggested `data-api-reporter` - however given it could be worth extending this in the future (e.g. so that it integrates with [the `upgrade-go-azure-sdk` tool](https://github.com/hashicorp/terraform-provider-azurerm/tree/main/internal/tools/update-go-azure-sdk)) - I suspect it’s probably worth leaving this as `differ` for the moment?


mbfrahry (approved):
LGTM! Nothing is jumping out as a negative other than potential noise-iness around information we don't care about so it'll be fun to see what information we get out of this first pass and how we can tweak it

tombuildsstuff:
Yeah I'm anticipating that we'll change the output UI after having learned more about what we're after - the information being output is useful but a little unwieldily at the moment - but we can change that more easily once we've got some more examples.

PR #3461, opened 2023-12-07
Add containerservice 2024-02-01

timja:
This is a starting point for:
https://github.com/hashicorp/terraform-provider-azurerm/issues/23982

API I'm after:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/stable/2024-02-01/managedClusters.json#L7330-L7333

I'm a bit confused on how services work when there's multiple under them.
e.g. the `2023-10-15` is for fleet manager and not AKS but nothing seems to distinguish the multiple services

PR #3469, opened 2023-12-11
add containerservices@2023-06-02-preview

ms-henglu:


stephybun:
@ms-henglu #3461 is adding an even newer preview version, is there a reason to only upgrade to `2023-06-02-preview`?

ms-henglu:
Hi @stephybun ,

Thank for the code review!

The background is that AKS team wants us to use the stable API version, so we don't want to support more preview features(which are not in the stable API version).

The 2023-06-02-preview is the minimum api version that has all the stable features(that 2023-10-01 has).

aristosvo:
Hi @ms-henglu! Could you in any way explain who 'we' and 'us' are here? Is there consensus on whether we want none non-GA features here or some specific ones? I made the mistake to request a non-preview API version for `ContainerService`, but as that doesn't contain the preview features currently in the provider already.

Would be awesome to hear more about this (available in Slack if you need me) as an external contributor and have some clear suggestions to work with.

ms-henglu:
Hello @aristosvo ,

Yes, MS(both TF team and AKS service team) and HC have an agreement that no more new preview features will onboard azurerm provider. Because AKS service team want the azurerm provider to use the stable api version, and we have already onboarded some preview features that are not GAed yet, so we couldn't switch to the latest stable api-version 2023-10-15 directly.

aristosvo:
Do you have any advice for companies that would like to use some preview features and use them via Terraform? Should they fork, do we want to create an `azurerm_kubernetes_cluster_preview` resource: what is the view on that?

ms-henglu:
Hi @aristosvo , we'd recommend user to use https://registry.terraform.io/providers/Azure/azapi/latest/docs/resources/azapi_resource to manage the preview features, because it supports all the api-versions. And we'll keep improving the user experience for the `azapi` provider.

katbyte:
@aristosvo - to be clear/explicit, the service team has asked us to only use the stable APIs for AKS in the provider which presents a problem as we currently support a number of features only available in their preview APIs, some with no expected/panned GA date. This obviously would be a big breaking change and a reduction of functionality within the provider that is widely used as well limit the future features we could support. We have pushed back and are currently discussing with the service team how we can move forward - and until we figure out a path forward it was decided not to add any new preview features to the provider to make the delta between what we support and would have to remove/solve for if they continue to insist we only use the stable API.

aristosvo:
@katbyte Thanks a lot! I understand the reasoning of the AKS service team, but it would indeed limit the abilities to use a test environment with preview features in a similar setup as a production environment.

From my point of view it would make sense to feature-gate such preview features with an explicit `preview_features_kubernetes_enabled` flag or have a separate `azurerm_kubernetes_*_preview` group of resources, but that would require a lot of extra testing 😅🫣

katbyte:
> `preview_features_kubernetes_enabled` flag or have a separate `azurerm_kubernetes_*_preview`

those are both ideas being considered but they each have challenges especially the flag, as what happens/how do you handle it being turned off for an existing resource where the latter option means a lot of duplicate codepaths to maintain (and both mean importing two AKS SDKs)

stephybun (approved):

PR #3497, opened 2023-12-15
Data: Rest Api Specs - regenerating based on 526799e0e4b4422f71ce95574ef9d13b880eab0e

hc-github-team-tf-azure:
This PR is automatically generated based on the commit 526799e0e4b4422f71ce95574ef9d13b880eab0e

PR #3498, opened 2023-12-15
automation: fixing the checkout path

tombuildsstuff:
The Github Actions agents aren't setting `$TMPPATH` so let's just output this into `./pandora-from-main` for now


tombuildsstuff:
This is a minor nit that'll be pass/fail needed to fix the GHA in https://github.com/hashicorp/pandora/pull/3497 - skipping review.

PR #3501, opened 2023-12-15
automation: `data-api-differ` should run against PRs and not the target branch

tombuildsstuff:
Copied this from the wrong GHA - right now this runs from `main` against `main` which means [no changes are output](https://github.com/hashicorp/pandora/pull/3500#issuecomment-1857602479) - this fixes that.


tombuildsstuff:
Skipping review since this is a minor nit and unfortunately the GHA can only be tested once it's in `main` - the current result being run is being from from `main` so will be a false-positive.

PR #3502, opened 2023-12-15
automation: forcing the fetch of `origin/main`

tombuildsstuff:
This shouldn't be needed since the GHA is checking out all branches, but hey.


tombuildsstuff:
Skipping review since I've validated the GHA output looks good - once rebased #3500 (or a regeneration) should prove this works.

PR #3504, opened 2023-12-15
automation: flipping the diff ordering

tombuildsstuff:
Fixes https://github.com/hashicorp/pandora/pull/3500#issuecomment-1857674623


tombuildsstuff:
This should be the final nit to ensure the `data-api-differ` is outputting the correct diff.

PR #3506, opened 2023-12-15
automation: validating the V2 API Definitions when there are changes to `api-definitions`

tombuildsstuff:
The Data API Differ is now configured, so this should ensure everything is fully validated


stephybun (approved):
LGTM 👍

PR #3511, opened 2023-12-18
Config: add BotService@2021-05-01-preview

neil-yechenwei:
Hi @tombuildsstuff , I submitted PR to switch to go-azure-sdk for BotService to fix [the issue](https://github.com/hashicorp/terraform-provider-azurerm/pull/23129) as you [suggested](https://github.com/hashicorp/terraform-provider-azurerm/pull/23129#issuecomment-1845701681). But seems CI in this PR is failed. Could you take a look? Thanks.


tombuildsstuff:
Dependent on https://github.com/hashicorp/pandora/pull/3576

tombuildsstuff:
hey @neil-yechenwei 

Taking a look through here I'm not entirely sure what's going on with the Github Actions - but in pulling this down locally it appears there's an inconsistency in the upstream API Definition, which I've opened https://github.com/Azure/azure-rest-api-specs/pull/27351 to fix.

As a part of getting that fixed I've opened #3609 which both adds a Data Workaround to fix that issue, but also imports the latest Stable API Version for Bot Service (`2022-09-15`) which should unblock this one - and then https://github.com/hashicorp/pandora/pull/3611 will update the integer constants to use the display name.

However since #3609 and #3611 are going to resolve this and import a stable version of this API - I hope you don't mind but I'm going to close this in favour of #3609, once both are merged this should become available however.

Thanks!

PR #3556, opened 2024-01-02
Add `ContainerService/2023-10-01`

aristosvo:
To fix https://github.com/hashicorp/terraform-provider-azurerm/issues/13066 I'd like to have the latest API version for ContainerService!

Thanks a lot!

If it is preferred to have an old preview version (to support some preview features) it might be that #3469 is the right one to merge, but I'm not completely up to speed with these kind of things 🧪


mbfrahry (approved):
LGTM!

PR #3563, opened 2024-01-04
add appplatform@2023-11-01-preview

ms-henglu:


config/resource-manager.hcl
```diff
@@ -33,7 +33,7 @@ service "applicationinsights" {
 }
 service "appplatform" {
   name      = "AppPlatform"
-  available = ["2023-07-01-preview", "2023-09-01-preview", "2023-12-01"]
+  available = ["2023-07-01-preview", "2023-09-01-preview", "2023-12-01", "2023-11-01-preview"]
```
tombuildsstuff, line 36: We're currently importing the later Stable API version `2023-12-01` - why do we need to import an earlier Preview API version?

tombuildsstuff: Where a Stable API Version becomes available we should be looking to migrate to that - since earlier Preview API versions can be removed within 3-6 months of a Stable API version becoming available in some cases.

Given there's functionality in the Preview API which hasn't yet made it to the Stable API, I'd expect there's a new version of the Preview API coming which supports both the Stable functionality plus the new Preview functionality - would you be able to reach out to the Service Team to confirm? That'd mean we can adopt the Stable API version today (meaning that there isn't a risk of the provider stopping working randomly) and then add the Preview functionality once the newer API version is available.

tombuildsstuff (approved):

PR #3478, opened 2023-12-12
add workaround for #26678

ziyeqf:
adding workaround for https://github.com/hashicorp/go-azure-sdk/issues/735 ( https://github.com/Azure/azure-rest-api-specs/issues/26678 )
will convert to ready to review once the test pipeline pass

PR #3578, opened 2024-01-08
nit: `tools/importer-rest-api-specs` - fixing the workaround for 26678 from #3478

tombuildsstuff:
This should be set back into `Clusters` and not `Cluster`.


tombuildsstuff:
Since this is a one-character nit that's blocking changes in the `main` branch I'm going to skip review for this one, so that #3577 can be regenerated, which'll show the result of this.

PR #3586, opened 2024-01-09
`importer-rest-api-specs` - support resource generation for resources with a scoped ID segment

stephybun:
Currently we skip resource generation if a candidate resource has a scoped resource ID segment. This PR removes that check and flips the logic for setting the Create/Read/Update Property payload instead of erroring.

These changes are required to be able to generate the resource Chaos Studio Targets. This shouldn't result in changes to the existing Data API definitions.

### DIFF
--- a/tools/importer-rest-api-specs/components/schema/builder.go
+++ b/tools/importer-rest-api-specs/components/schema/builder.go
@@ -378,11 +378,10 @@ func (b Builder) findCreateUpdateReadPayloads(input resourcemanager.TerraformRes
 	out.createModelName = *createOperation.RequestObject.ReferenceName
 	out.createPayload = createModel
 	createPropsModelName, createPropsModel := out.getPropertiesModelWithinModel(out.createPayload, b.models)
-	if createPropsModelName == nil || createPropsModel == nil {
-		return nil, fmt.Errorf("couldn't find `Properties` model for Create Payload")
+	if createPropsModelName != nil || createPropsModel != nil {
+		out.createPropertiesPayload = *createPropsModel
+		out.createPropertiesModelName = *createPropsModelName
 	}
-	out.createPropertiesPayload = *createPropsModel
-	out.createPropertiesModelName = *createPropsModelName
 
 	// Read has to exist
 	readOperation, ok := b.operations[input.ReadMethod.MethodName]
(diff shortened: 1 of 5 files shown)


tools/importer-rest-api-specs/components/schema/builder.go
```diff
@@ -378,11 +378,10 @@ func (b Builder) findCreateUpdateReadPayloads(input resourcemanager.TerraformRes
 	out.createModelName = *createOperation.RequestObject.ReferenceName
 	out.createPayload = createModel
 	createPropsModelName, createPropsModel := out.getPropertiesModelWithinModel(out.createPayload, b.models)
-	if createPropsModelName == nil || createPropsModel == nil {
-		return nil, fmt.Errorf("couldn't find `Properties` model for Create Payload")
+	if createPropsModelName != nil || createPropsModel != nil {
+		out.createPropertiesPayload = *createPropsModel
+		out.createPropertiesModelName = *createPropsModelName
```
tombuildsstuff, line 383: As both `ModelName` and `Payload` for the `Properties` object are becoming optional rather than required - we should [update the `operationPayloads` struct to make these pointers to signify that](https://github.com/hashicorp/pandora/blob/19980c54093303a2d500de7e9b84c15c4b82a27e/tools/importer-rest-api-specs/components/schema/internal.go#L7-L22) - which'd make this more obvious that these are optional in other parts of the codebase

tools/importer-rest-api-specs/components/schema/builder.go
```diff
@@ -401,11 +400,10 @@ func (b Builder) findCreateUpdateReadPayloads(input resourcemanager.TerraformRes
 	out.readPayload = readModel
 	// then find the `Properties` model within this
 	readPropsModelName, readPropsModel := out.getPropertiesModelWithinModel(out.readPayload, b.models)
-	if readPropsModelName == nil || readPropsModel == nil {
-		return nil, fmt.Errorf("couldn't find `Properties` model for Read Payload")
+	if readPropsModelName != nil || readPropsModel != nil {
+		out.readPropertiesModelName = *readPropsModelName
+		out.readPropertiesPayload = *readPropsModel
 	}
```
tombuildsstuff, line 406: (as above)

tombuildsstuff (approved):
Given the `properties` fields are becoming optional rather than required, we should update the `operationPayloads` object to make these pointers - but if we can update that then this otherwise LGTM 👍

tools/importer-rest-api-specs/components/schema/builder.go
stephybun: As discussed this change causes unwanted effects in the existing generated resources, details in #3588

tools/importer-rest-api-specs/components/schema/builder.go
stephybun: As discussed this change causes unwanted effects in the existing generated resources, details in #3588

PR #3587, opened 2024-01-09
tools/generator-go-sdk: fixing an issue where Response rather than `HttpResponse` was being used

tombuildsstuff:
Two different code paths are used depending on the Response Object shape, when a Model or a List of a Model is returned we include predicate filtering - otherwise (e.g. for a Basic Type [string/int/etc]) we don't however we only tested one of these code paths previously - so this PR fixes that.

Ultimately this fixes the compilation failure seen in https://github.com/hashicorp/go-azure-sdk/pull/799

This PR also updates the SDK to use `hashicorp/go-retryablehttp` to workaround an issue when requests randomly fail (which appears to be due to security tooling).


tools/importer-msgraph-metadata/go.mod
```diff
@@ -71,3 +71,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/hashicorp/pandora/tools/sdk => ../sdk
```
stephybun, line 75: Are we supposed to be doing this replacement in main as well?

stephybun (approved):
Left one question, otherwise this LGTM 👍

tools/importer-msgraph-metadata/go.mod
tombuildsstuff: Yep this is intentional - it ensures the `importer-msgraph-metadata` tool aligns with the other tools which already include the replace - which means that the `sdk` directory is sourced from the local file system, rather than [this commit (vendored as `v0.0.0-20230809001200-97c549958463` in the `go.mod`)](https://github.com/hashicorp/pandora/tree/97c5499584637328cfda7ee169668f9286fa6859) which it's doing today.

PR #3596, opened 2024-01-10
automation: updating the Terraform Generator to use Data API V2

tombuildsstuff:
This PR also updates the way the `data-api` tool is built and launched to match the way the `generator-go-sdk` and `generator-terraform` tools are built/launched (that is, installing them onto the `$GOPATH`) - meaning that the shell scripts contain all of the dependencies needed to launch the Data API v2. This means that the process is now fully reproducible with a fresh checkout, which makes debugging issues simpler.

Notably as a part of doing this the `wrapper-automation` tool now takes the path to the API Definitions directory, matching the behaviour of Data API V2.

Finally this PR cleans up the Github Action for the Go SDK Generator, so that we no longer build V1 of the Data API (since this is no longer using it).

### DIFF
--- a/.github/workflows/automation-regenerate-go-sdk.yaml
+++ b/.github/workflows/automation-regenerate-go-sdk.yaml
@@ -24,16 +24,6 @@ jobs:
         with:
           go-version: '1.21.3'
 
-      - name: Setup .NET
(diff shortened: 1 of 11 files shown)


stephybun (approved):
LGTM 👍

Issue #3583, opened 2024-01-09
Data API V2: Remove old Data API


stephybun:
Once Terraform Resource generation has been switched on to use Data API V2, we should begin the process of decommissioning and removing the old Data API from the project.

1. ~Disable generation of the legacy definitions in the scripts and GHA (allow the pipeline to run for 1-2 weeks before continuing the actions below)~ done
2. ~Disable the `tools/extract-tf-resource-ids` GHA job~ done
3. ~Remove the logic and templates that output the legacy definitions in the `tools/importer-rest-api-specs` https://github.com/hashicorp/pandora/pull/3633~ done
4. ~Remove support from `wrapper-automation` https://github.com/hashicorp/pandora/pull/3666~ done
5. Remove the `tools/extract-tf-resource-ids` tool (and documentation references)
6. Remove the `data/` folder

PR #3600, opened 2024-01-10
`tools/importer-rest-api-specs`: disabling the generation of the V1 API Definitions

tombuildsstuff:
Also disabling the `extract-tf-resource-ids` job since this is now superfluous

The first half of #3583


mbfrahry (approved):
LGTM!

tombuildsstuff:
Opting to skip the `microsoft-graph` failure here since this is going to be reworked/isn't critical.

PR #3609, opened 2024-01-11
`tools/importer-rest-api-specs`: adding a data workaround for `Azure/azure-rest-api-specs#27351` / importing `BotService` @ `2022-09-15`

tombuildsstuff:
This PR adds a data workaround for https://github.com/Azure/azure-rest-api-specs/pull/27351 which enables importing `BotService` @ `2022-09-15`.

The Swagger defines the same Path with different types of segments depending on the HTTP Method, where the DELETE and GET operations use a String segment - but the PATCH and POST operations use a Constant segment - meaning that whilst we detect the Common ID (using the Constant segments) it's inconsistent).

Related: https://github.com/hashicorp/go-azure-helpers/pull/203

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -71,6 +71,10 @@ service "blueprint" {
   name      = "Blueprints"
   available = ["2018-11-01-preview"]
 }
+service "botservice" {
+  name      = "BotService"
+  available = ["2022-09-15"]
+}
 service "chaos" {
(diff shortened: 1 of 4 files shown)


stephybun (approved):
LGTM 👍

tombuildsstuff:
Related https://github.com/hashicorp/pandora/pull/3611

Issue #3010, opened 2023-09-04
Constants: pulling the display name from `x-ms-enum`


tombuildsstuff:
Some Swaggers define a display name for integer (and potentially float) constants - as such we should look to use this for these two constant types:

https://github.com/Azure/azure-rest-api-specs/blob/80c21c17b4a7aa57f637ee594f7cfd653255a7e0/specification/botservice/resource-manager/Microsoft.BotService/stable/2022-09-15/botservice.json#L2187-L2210

Note: we shouldn't do this for string constants, since the string value will be a good enough identifier

PR #3611, opened 2024-01-11
`tools/importer-rest-api-specs`: support for pulling the Display Name for an Integer Constant if defined

tombuildsstuff:
This commit fixes https://github.com/hashicorp/pandora/issues/3010 by checking for a Display Name within the `x-ms-enum` block and pulling from that if a (Display) Name is defined for an Integer constant.

Whilst I considered doing this for Float and String constants too, I've intentionally omitted those since Float constants are rare (so we can apply this same fix if needed down the line) - and for String constants there's little benefit to changing at this point, since we normalize a key from the value for the constant.

Ultimately this should ensure that when we import the constant `EmailChannelAuthMethod` for `BotService` that we output this as:

> `Password`: `0`
> `Graph`: `1`

Which is far more descriptive than:

> `Zero`: `0`
> `One`: `1`

---

There's a couple of changes to the existing data, but it's nothing major:

```diff
diff --git a/api-definitions/resource-manager/CosmosDB/2022-11-15/Mongorbacs/Constant-MongoRoleDefinitionType.json b/api-definitions/resource-manager/CosmosDB/2022-11-15/Mongorbacs/Constant-MongoRoleDefinitionType.json
index 6283a51ce7..0c3a0f9cf9 100644
--- a/api-definitions/resource-manager/CosmosDB/2022-11-15/Mongorbacs/Constant-MongoRoleDefinitionType.json
+++ b/api-definitions/resource-manager/CosmosDB/2022-11-15/Mongorbacs/Constant-MongoRoleDefinitionType.json
@@ -3,12 +3,12 @@
        "type": "Integer",
        "values": [
                {
-                       "key": "One",
-                       "value": "1"
+                       "key": "BuiltInRole",
+                       "value": "0"
                },
                {
-                       "key": "Zero",
-                       "value": "0"
+                       "key": "CustomRole",
+                       "value": "1"
                }
        ]
 }
\ No newline at end of file
diff --git a/api-definitions/resource-manager/MixedReality/2021-01-01/Key/Constant-Serial.json b/api-definitions/resource-manager/MixedReality/2021-01-01/Key/Constant-Serial.json
index f21186275a..1e385dd515 100644
--- a/api-definitions/resource-manager/MixedReality/2021-01-01/Key/Constant-Serial.json
+++ b/api-definitions/resource-manager/MixedReality/2021-01-01/Key/Constant-Serial.json
@@ -3,11 +3,11 @@
        "type": "Integer",
        "values": [
                {
-                       "key": "One",
+                       "key": "Primary",
                        "value": "1"
                },
                {
-                       "key": "Two",
+                       "key": "Secondary",
                        "value": "2"
                }
        ]
```

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/constants/interface.go
+++ b/tools/importer-rest-api-specs/components/parser/constants/interface.go
@@ -15,7 +15,13 @@ import (
 )
 
 type constantExtension struct {
+	// name defines the Name that should be used for this Constant
 	name string
+
+	// valuesToDisplayNames defines any display name overrides that should be used for this Constant
+	// NOTE: whilst the API Definitions may define a value with no display name - this map contains
+	// only values with a name defined.
+	valuesToDisplayNames *map[interface{}]string
 }
 
 type ParsedConstant struct {
@@ -96,6 +102,14 @@ func MapConstant(typeVal spec.StringOrArray, fieldName string, values []interfac
 			}
 
 			key := keyValueForInteger(int64(value))
+			// if an override name is defined for this Constant then we should use it
+			if constExtension.valuesToDisplayNames != nil {
+				overrideName, hasOverride := (*constExtension.valuesToDisplayNames)[value]
+				if hasOverride {
+					key = overrideName
+				}
+			}
+
 			val := fmt.Sprintf("%d", int64(value))
 			normalizedName := normalizeConstantKey(key)
 			keysAndValues[normalizedName] = val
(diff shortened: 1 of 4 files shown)


tools/importer-rest-api-specs/components/parser/constants/interface.go
```diff
@@ -145,13 +159,38 @@ func parseConstantExtensionFromExtension(field spec.Extensions) (*constantExtens
...
 		if strings.EqualFold(k, "name") {
 			normalizedEnumName := cleanup.NormalizeName(v.(string))
 			enumName = &normalizedEnumName
 		}
 
+		if strings.EqualFold(k, "values") {
+			items := v.([]interface{})
+			displayNameOverrides := make(map[interface{}]string)
+			for _, itemRaw := range items {
+				item := itemRaw.(map[string]interface{})
```
mbfrahry, line 174: Do we want to check that this isn't nil or is it fine?

mbfrahry (approved):
LGTM with a minor comment

tools/importer-rest-api-specs/components/parser/constants/interface.go
tombuildsstuff: The extension is defined in this manner, and everything we're importing matches it - so I think this is fine as-is for now. If this is an issue we can fix it at that point, but that's likely going to be bad data - so would want resolving upstream anyway - as such this is fine 👍

PR #3614, opened 2024-01-12
identity: fix for `DelegatedResources` field

### LINKED ISSUE #3612: Add new identity property: `DelegatedResources`
I'm trying to use the [Containers/2023-06-02-preview SDK](https://github.com/hashicorp/pandora/pull/3469) from the hashicorp/go-azure-sdk , but the identity part of cluster contains some brand new property: DelegatedResources

```go
type ManagedClusterIdentity struct {
	DelegatedResources     *map[string]DelegatedResource                `json:"delegatedResources,omitempty"`
	PrincipalId            *string                                      `json:"principalId,omitempty"`
	TenantId               *string                                      `json:"tenantId,omitempty"`
	Type                   *ResourceIdentityType                        `json:"type,omitempty"`
	UserAssignedIdentities *map[string]UserAssignedIdentitiesProperties `json:"userAssignedIdentities,omitempty"`
}
```

This causes a mismatch in the generator, which makes usage of the `github.com/hashicorp/go-azure-helpers/resourcemanager/identity` package impossible

aristosvo:
Fixes #3612

## TODO:
- [x] Tests


tools/importer-rest-api-specs/components/parser/commonschema/custom_field_identity_system_or_user_assigned_map.go
```diff
@@ -107,6 +107,10 @@ func (systemOrUserAssignedIdentityMapMatcher) IsMatch(_ models.FieldDetails, def
 			continue
 		}
 
+		if strings.EqualFold(fieldName, "DelegatedResources") {
```
tombuildsstuff, line 110: can we add a little context for this:

```suggestion
		// Per the API Definition, the `DelegatedResources` field is for `internal use only` - therefore we should ignore this if it's present
		if strings.EqualFold(fieldName, "DelegatedResources") {
```

tombuildsstuff (approved):
LGTM 👍

tombuildsstuff:
`Detect changes to the API Definitions` can be ignored, we'll disable that from these PRs in a separate commit.

PR #3617, opened 2024-01-12
automation: disabling `Detect changes to the API Definitions` when `importer-rest-api-specs` changes

tombuildsstuff:
The API definitions aren't changed at this point (but might make sense for us to handle in the end-to-end test) as such I'm disabling this for now since this fails for forked PRs.


mbfrahry (approved):
LGTM

PR #3620, opened 2024-01-12
`tools/generator-go-sdk`: updating `LoadTestService` and `ManagedIdentity` to use the new base layer

tombuildsstuff:
This also requires updating `generator-terraform` to account to account for that.

However - given these updates need to be coordinated - merging this change will (temporarily) break the `generator-terraform` automation, as the resulting code won't compile. This is because the SDK version vendored into `hashicorp/terraform-provider-azurerm` still uses the old base layer, as such the Go SDK will need to be released and then updated in `hashicorp/terraform-provider-azurerm` by hand - at which point the automation will work as usual.

Whilst it's possible to stagger the updates such that we released the Go SDK switching these two services over to the new base layer, and then did the same for the Terraform Provider - the end result is the same insofar as we'd end up needing to manually reconcile these - so I think the "least bad" thing to do for now is to keep this in a single PR and then roll this out.

---

I've intentionally opted to _not_ move `ContainerService` in this PR due to the size of the surface area/other open PR's - but we'll want to coordinate that one in time.

### DIFF
--- a/tools/generator-go-sdk/main.go
+++ b/tools/generator-go-sdk/main.go
@@ -33,8 +33,6 @@ func main() {
 		// with services already used in `terraform-provider-azurerm`. These services will be gradually removed
 		// from this list to ensure they're migrated across to using `hashicorp/go-azure-sdk`s base layer.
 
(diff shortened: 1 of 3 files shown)


mbfrahry (approved):
LGTM

PR #3621, opened 2024-01-14
Add support for billing service

daniel-edwards-nz:
Adding support for the [billing service APIs](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/billing/resource-manager/readme.md) as I'd like to help develop and use these in the AzureRM terraform provider.

Included the two stable versions "2020-05-01" and "2021-10-01". Also included "2019-10-01-preview" as there are quite a few things in here which haven't yet made it to stable which would be useful.

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -68,7 +68,7 @@ service "batch" {
   available = ["2023-05-01", "2023-11-01"]
 }
 service "billing" {
-  name      = "billing"
+  name      = "Billing"
   available = ["2019-10-01-preview", "2020-05-01", "2021-10-01"]
 }
 service "blueprint" {
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -67,6 +67,10 @@ service "batch" {
   name      = "Batch"
   available = ["2023-05-01", "2023-11-01"]
 }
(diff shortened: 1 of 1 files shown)


config/resource-manager.hcl
```diff
@@ -67,6 +67,10 @@ service "batch" {
   name      = "Batch"
   available = ["2023-05-01", "2023-11-01"]
 }
+service "billing" {
+  name      = "billing"
```
tombuildsstuff, line 71: This wants to be Title Cased, which is why the unit tests are failing - mind updating this?

```suggestion
  name      = "Billing"
```

tombuildsstuff (changes requested):
hey @daniel-edwards-nz 

Thanks for this PR.

I've taken a look through and left one comment inline to fix the test, but if we can fix that up then this otherwise LGTM 👍 

Thanks!

config/resource-manager.hcl
daniel-edwards-nz: Oops, thanks for catching the casing issue @tombuildsstuff. I've added the suggested update now so hopefully those tests should pass and this will be good to go!

tombuildsstuff (approved):
LGTM 👍

PR #3628, opened 2024-01-16
generator-terraform: nil check ModelToModel mappings

stephybun:
Fixes https://github.com/hashicorp/pandora/actions/runs/7538602136/job/20519702891

PR #3631, opened 2024-01-16
tools/data-api: fixing the mapping of Terraform specific objects

tombuildsstuff:
This updates the mapping logic so that we return a pointer to a map as the result of the function, rather than a map directly - this allows us to differentiate between unset (e.g. an error has occurred) and when a value exists - which fixes the root cause of https://github.com/hashicorp/pandora/pull/3628 (and means that downstream tooling can avoid nil-checking these)

This PR also adds some additional mappings for the `repository` types to the types used for the API responses rather than string casting - since string casting in Go will map the cast the literal value (which'll always be valid) meaning that we wouldn't necessarily map the value we're defining across.

### DIFF
--- a/tools/data-api/internal/endpoints/v1/terraform.go
+++ b/tools/data-api/internal/endpoints/v1/terraform.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/go-chi/render"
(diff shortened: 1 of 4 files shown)


mbfrahry (approved):
LGTM

PR #3636, opened 2024-01-17
tools/importer-rest-api-specs: extending the workaround for https://github.com/Azure/azure-rest-api-specs/pull/27351 to account for the constant being incorrectly defined

tombuildsstuff:
Fixes https://github.com/hashicorp/pandora/pull/3635/files#diff-dbffa561a5c7eb7b00c278bb4fcdc7fc2098a3b70bf572d6f119d32dc9fd1aaaR3

This should be explicitly defined as an integer, since `number` (without an explicit `format`) means float.


jackofallops (approved):
LGTM 👍

Issue #3639, opened 2024-01-17
`tools/generator-go-sdk`: unmarshaling a list of discriminated types is broken


tombuildsstuff:
When switching `DataShare` over to the new base layer, the `ListByShare` operation is broken, returning `map[string]interface{}` as the type of `SynchronizationSetting`, rather than the correctly unmarshalled object.

This is accounted for/works fine when using the old base layer - as such this is scoped to the new base layer.

Working (old base layer): https://github.com/hashicorp/go-azure-sdk/commit/a476e69ed7112cd5112ffd0b71b0405a1cbe1c8c#diff-b7ab6a17ff3638f8b2855cc7cbce18daea707f6b7b7b68f523a1767abc94e0b1L117-L126

```go
// responderForListByShare handles the response to the ListByShare request. The method always
// closes the http.Response Body.
func (c SynchronizationSettingClient) responderForListByShare(resp *http.Response) (result ListByShareOperationResponse, err error) {
	type page struct {
		Values   []json.RawMessage `json:"value"`
		NextLink *string           `json:"nextLink"`
	}
	var respObj page
	err = autorest.Respond(
		resp,
		azure.WithErrorUnlessStatusCode(http.StatusOK),
		autorest.ByUnmarshallingJSON(&respObj),
		autorest.ByClosing())
	result.HttpResponse = resp
	temp := make([]SynchronizationSetting, 0)
	for i, v := range respObj.Values {
		val, err := unmarshalSynchronizationSettingImplementation(v)
		if err != nil {
			err = fmt.Errorf("unmarshalling item %d for SynchronizationSetting (%q): %+v", i, v, err)
			return result, err
		}
		temp = append(temp, val)
	}
	result.Model = &temp
	result.nextLink = respObj.NextLink
	if respObj.NextLink != nil {
		result.nextPageFunc = func(ctx context.Context, nextLink string) (result ListByShareOperationResponse, err error) {
			req, err := c.preparerForListByShareWithNextLink(ctx, nextLink)
			if err != nil {
				err = autorest.NewErrorWithError(err, "synchronizationsetting.SynchronizationSettingClient", "ListByShare", nil, "Failure preparing request")
				return
			}

			result.HttpResponse, err = c.Client.Send(req, azure.DoRetryWithRegistration(c.Client))
			if err != nil {
				err = autorest.NewErrorWithError(err, "synchronizationsetting.SynchronizationSettingClient", "ListByShare", result.HttpResponse, "Failure sending request")
[...]

PR #3640, opened 2024-01-17
`tools/generator-go-sdk`: fixing support for top-level discriminated lists in the new base layer

tombuildsstuff:
This PR fixes an issue in the new base layer where a list of Discriminated Types is returned from a (in this case, List) operation.

This mirrors the behaviour of the old base layer, by creating an intermediate slice to deserialize into, calling `unmarshal` on each of the items.

Fixes https://github.com/hashicorp/pandora/issues/3639
Unblocks https://github.com/hashicorp/terraform-provider-azurerm/pull/24481


jackofallops (approved):
LGTM 👍

PR #3642, opened 2024-01-17
`tools/data-api`: consolidating the `microsoft-graph` endpoints together

tombuildsstuff:
Chatting with @manicminer - rather than having two separate Data Sources for `Beta` and `Stable V1` for Microsoft Graph, we're going to try and use the same versioning approach that Resource Manager does - that is `/v1/{DataSourceType}/{ServiceName}/{ApiVersion}/{ApiResource}` - e.g. `/v1/microsoft-graph/Applications/{beta|v1stable}/Application`.

As such this PR:

1. Consolidates `/v1/microsoft-graph/beta` and `/v1/microsoft-graph/stable-v1` into `/microsoft-graph`.
2. Updates the constants to reflect this.
3. Removes the unused `/infrastructure` endpoint (the SDK uses `/v1/health` rather than `/infrastructure/health`, so this is fine)

### DIFF
--- a/tools/data-api/internal/endpoints/routing.go
+++ b/tools/data-api/internal/endpoints/routing.go
@@ -12,23 +12,10 @@ import (
 func Router(directory string, serviceNames *[]string) func(chi.Router) {
 	return func(router chi.Router) {
 		router.Route("/v1", infrastructure.Router)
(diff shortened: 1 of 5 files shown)


stephybun (approved):
LGTM 👍

PR #3647, opened 2024-01-18
commonids - add common IDs for chaos studio capability and target

stephybun:
This PR adds a common ID for Chaos Studio Capability and Target and also fixes the swapping out of test data for fields that have been renamed, as well as omits generation of the complete test configuration if all fields are Required.

Dependent on https://github.com/hashicorp/go-azure-sdk/pull/819

### DIFF
--- a/tools/importer-rest-api-specs/components/testing/attribute_value_for_field.go
+++ b/tools/importer-rest-api-specs/components/testing/attribute_value_for_field.go
@@ -147,6 +147,13 @@ var attributeValuesForBasicTypes = map[resourcemanager.TerraformSchemaFieldType]
 			return &val, nil
 		}
 
+		for k, v := range testData.Strings {
+			if strings.EqualFold(field.HclName, k) {
+				out := hclwrite.TokensForValue(cty.StringVal(v))
+				return &out, nil
+			}
+		}
+
 		if strings.EqualFold(field.HclName, "description") {
 			// Description for the Example Thing
 			out := hclwrite.TokensForValue(cty.StringVal(fmt.Sprintf("Description for the %s", resourceDisplayName)))
--- a/tools/importer-rest-api-specs/components/testing/attribute_value_for_field.go
+++ b/tools/importer-rest-api-specs/components/testing/attribute_value_for_field.go
@@ -148,7 +148,7 @@ var attributeValuesForBasicTypes = map[resourcemanager.TerraformSchemaFieldType]
 		}
 
 		for k, v := range testData.Strings {
-			if strings.EqualFold(field.HclName, k) {
+			if field.HclName == k {
 				out := hclwrite.TokensForValue(cty.StringVal(v))
 				return &out, nil
 			}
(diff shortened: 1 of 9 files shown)


tools/importer-rest-api-specs/components/testing/complete.go
```diff
@@ -13,6 +13,19 @@ func (tb TestBuilder) generateCompleteTest(dependencies *testDependencies) (*str
...
+
+	completeRequired := false
+	for _, details := range topLevelModel.Fields {
+		if details.Required == false {
+			completeRequired = true
+			break
+		}
+	}
+
+	if !completeRequired {
```
stephybun, line 25: Returning `nil, nil` here since in `generate.go` we have checks to omit generation of the complete test if this function returns nil: 

https://github.com/hashicorp/pandora/blob/a608cd5ff33fb038760d51d246c8d917779f02cc/tools/importer-rest-api-specs/components/testing/generate.go#L60-L63

However this unit test fails expects that if there are no properties that we generate an empty config which I don't think is right. Is there a reason that we've written this test case like so?
https://github.com/hashicorp/pandora/blob/a608cd5ff33fb038760d51d246c8d917779f02cc/tools/importer-rest-api-specs/components/testing/complete_test.go#L13

tombuildsstuff: > Is there a reason that we've written this test case like so?

Looking at the code I'm gonna wager the Basic tests were copied and then updated to match - I don't think there's a need to support that scenario now since we're filtering the tests out - so we should be good to remove the test 👍

It's worth noting (given the name) that there _is_ a scenario where the `properties` object is required to be sent (but doesn't contain any fields - since else the API crashes [we've seen this a handful of times]) - but that's not what this test is testing, so it's fine 👍

tools/importer-rest-api-specs/components/testing/dependencies.go
```diff
@@ -171,7 +171,7 @@ func DetermineDependencies(field, providerPrefix string, dependencies *testDepen
 		"subscription_id":               {dependencies.setNeedsClientConfig, fmt.Sprintf("data.%s_client_config.test.subscription_id", providerPrefix)},
 		// Currently only Chaos Studio Targets has this property which can accept many different resource IDs
 		// for now setting this to storage is sufficient, but we will need to consider how to proceed in future
-		"target_resource_id":        {dependencies.setNeedsStorageAccount, fmt.Sprintf("%s_storage_account.test.id", providerPrefix)},
+		"target_resource_id":        {dependencies.setNeedsKubernetesCluster, fmt.Sprintf("%s_kubernetes_cluster.test.id", providerPrefix)},
```
tombuildsstuff, line 174: _nit_ can we update the comment to mention Kubernetes Cluster and not Storage.

tools/importer-rest-api-specs/components/testing/attribute_value_for_field.go
```diff
@@ -147,6 +147,13 @@ var attributeValuesForBasicTypes = map[resourcemanager.TerraformSchemaFieldType]
 			return &val, nil
 		}
 
+		for k, v := range testData.Strings {
+			if strings.EqualFold(field.HclName, k) {
```
tombuildsstuff, line 151: Double-checking: do we need the `EqualFold` here? At this point the input should be normalised (so if it's not then we've got a problem elsewhere)?

tombuildsstuff (approved):
Left a couple of minor things but this otherwise LGTM 👍

PR #3648, opened 2024-01-18
`tools/generator` - fix generator test

catriona-m:


tombuildsstuff (approved):
LGTM 👍

Aside: there aren't tests covering the ID Parser Tests (which confused me for a second reviewing this) - we should probably look to add some covering that?

PR #3651, opened 2024-01-18
importer-rest-api-specs - nil check reference name when determining nested optional fields

stephybun:
Fixes https://github.com/hashicorp/pandora/actions/runs/7571557477/job/20619333989

### DIFF
--- a/tools/importer-rest-api-specs/components/testing/complete.go
+++ b/tools/importer-rest-api-specs/components/testing/complete.go
@@ -49,7 +49,7 @@ func (tb TestBuilder) generateCompleteTest(dependencies *testDependencies) (*str
 func hasOptionalField(input map[string]resourcemanager.TerraformSchemaFieldDefinition, schemaModels map[string]resourcemanager.TerraformSchemaModelDefinition) bool {
 	for _, details := range input {
 		if details.ObjectDefinition.NestedObject != nil {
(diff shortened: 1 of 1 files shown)


tools/importer-rest-api-specs/components/testing/complete.go
```diff
@@ -49,9 +49,11 @@ func (tb TestBuilder) generateCompleteTest(dependencies *testDependencies) (*str
 func hasOptionalField(input map[string]resourcemanager.TerraformSchemaFieldDefinition, schemaModels map[string]resourcemanager.TerraformSchemaModelDefinition) bool {
 	for _, details := range input {
 		if details.ObjectDefinition.NestedObject != nil {
-			nestedModel := schemaModels[*details.ObjectDefinition.NestedObject.ReferenceName]
-			if hasOptionalField(nestedModel.Fields, schemaModels) {
-				return true
+			if details.ObjectDefinition.NestedObject.ReferenceName != nil {
```
tombuildsstuff, line 52: _minor_ but the rest of the codebase checks based on the Type of the (Nested) Object Definition (and if Type is a Reference then we require a ReferenceName) - so it'd be good to match that behaviour:

```suggestion
			if details.ObjectDefinition.NestedObject.Type == resourcemanager.ReferenceApiObjectDefinitionType {
```

tombuildsstuff (approved):
One minor comment for behavioural consistency with the rest of the codebase - but otherwise 👍

PR #3664, opened 2024-01-22
`tools/generator-go-sdk`: conditionally calling the unmarshal function for discriminated types

tombuildsstuff:
When the Response Object contains a Discriminated Type we need to determine whether it's a Parent or an Implementation, so that we can determine whether or not we need to call the Parent's unmarshal function.

This fixes an issue spotted in https://github.com/hashicorp/go-azure-sdk/pull/822 where the Unmarshal method was always called - which wouldn't work in the event of a Discriminated Implementation being used as the Response Object.


jackofallops (approved):
LGTM 👍

PR #3673, opened 2024-01-22
`keyvault`: `2023-07-01` uses the new base layer / removing unused API Versions

tombuildsstuff:
This PR:

* Removes the now unused API Versions `2021-10-01` and `2022-11-01` for `KeyVault`
* Updates `KeyVault` API Version `2023-07-01` to use the new base layer

---

In creating an up-to-date repro for the Key Vault Constant casing issues, it appears that `2023-07-01` doesn't contain these issues - and whilst it's possible that earlier versions have been fixed - for the sake of a migration path this PR intentionally leaves `2023-02-01` on the older base layer - whilst allowing us a path to fixing some of the other open issues such as https://github.com/hashicorp/terraform-provider-azurerm/pull/24449

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -280,7 +280,7 @@ service "iotcentral" {
 }
 service "keyvault" {
   name      = "KeyVault"
-  available = ["2021-10-01", "2022-11-01", "2023-02-01", "2023-07-01"]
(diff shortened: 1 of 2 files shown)


stephybun (approved):
LGTM 👍

PR #3681, opened 2024-01-23
`config`: removing unused older API Versions

tombuildsstuff:
This PR removes older API versions which are no longer used in `hashicorp/terraform-provider-azurerm`.


stephybun (approved):
LGTM 👍

PR #3691, opened 2024-01-24
`tools/generator-go-sdk`: skipping unmarshalling the response object for LROs

tombuildsstuff:
Depending on the Operation and status code being returned, whilst a response object may be defined, it's not necessarily matching the initial http response - but could well be matching the LRO result.

As such this PR fixes https://github.com/hashicorp/go-azure-sdk/issues/839 by no longer unmarshalling the Response Object when the Operation is a LRO.

We may want to reevaluate this in the future, since the `XXXThenPoll` methods should _arguably_ be unmarshalling the LRO Response - but given the payload may differ between the initial HTTP Response and the LRO Response, that may need to be done through adding an `UnmarshalResponse` function, due to the complexity of unmashalling discriminated types.


jackofallops (approved):
LGTM 👍

PR #3711, opened 2024-01-26
tooling: updating to Go 1.21.6

tombuildsstuff:
This also switches to using Go Version files, so that we can make this change in 2 places rather than a bunch.


manicminer (approved):
LGTM

PR #3713, opened 2024-01-26
`tools/importer-rest-api-specs`: support for importing `Resources/*/Tags`

tombuildsstuff:
This PR unblocks migrating the `Tags` client from `Azure/azure-sdk-for-go` to `hashicorp/go-azure-sdk` as used here: https://github.com/hashicorp/terraform-provider-azurerm/blob/75182167d45fbfd0d68b0c9e9ce6340496276f0b/internal/services/resource/client/client.go#L121-L126

We're currently skipping importing both the `AzureFirewallFqdnTags` and the Consumption tags - taking a quick look at the generated output for both I'm not necessarily sure we should be supporting these, so I've added exceptions for these back in for the moment.


mbfrahry (approved):
LGTM

PR #3720, opened 2024-01-30
`tools/importer-rest-api-specs`: supporting `AllOf` containing only a description

tombuildsstuff:
`AllOf` is intended to be a reference, however sometimes can contain a description instead.


jackofallops (approved):
LGTM 👍

Issue #2848, opened 2023-07-28
`HybridAKS` @ `2022-09-01-preview` - eventual consistency/swagger issue


tombuildsstuff:
```
2023/07/28 07:01:46 [DEBUG] No Data API Endpoint is specified - using local-only mode
2023-07-28T07:01:46.891+0200 [INFO]  Finding only the Services "HybridAKS"
2023-07-28T07:01:49.221+0200 [INFO]  Importer for Service "HybridAKS".Importer for API Version "2022-09-01-preview": ❌ Service "HybridAKS" - Api Version "2022-09-01-preview"
2023-07-28T07:01:49.221+0200 [ERROR] Importer for Service "HybridAKS".Importer for API Version "2022-09-01-preview":      💥 Error: parsing Swagger files: parsing files in "../../swagger/specification/hybridaks/resource-manager/Microsoft.HybridContainerService/preview/2022-09-01-preview": parsing definition: finding resources for tag "HybridAKS": finding nested items yet to be parsed: appending model: appending models: different model objects for Model "provisionedClustersCommonProperties": first had 1 fields but second had 4 fields.

First fields: map[arcAgentProfile:{Required:false ReadOnly:false Sensitive:false JsonName:arcAgentProfile Description:Arc agentry configuration for the provisioned cluster. CustomFieldType:<nil> ObjectDefinition:Type: "Reference" / Reference: "ArcAgentProfile" / Minimum: 0 / Maximum: 0 / Unique Items: false / Nested Item: "<nil>"}].

[...]

PR #3721, opened 2024-01-30
`hybridaks`: importing API Version `2024-01-01`

tombuildsstuff:
As @wuxu92 has mentioned in #2848, it appears that the issue blocking `HybridAKS` has been resolved - however we also need to detect the segments `connectedClusterResourceUri` and `customLocationResourceUri` as scope's.

As such this PR adds support for `HybridAzureKubernetesService` (which is intentionally spelled out since we also have `HybridKubernetes`) @ `2024-01-01`.


mbfrahry (approved):
LGTM!

PR #3737, opened 2024-01-31
Cleanup: removing V1 of the Data API

tombuildsstuff:
This PR removes V1 of the Data API by removing `./data` and `./tools/extract-tf-resource-ids` - and a few minor nits (automation, readme)

Since the Microsoft Graph importer hasn't yet been updated to use Data API v2, this PR removes the `Verify All Microsoft Graph Services can be Imported and Generated` job for the moment.

Given the size of the removals in `./data` - the first commit (and Github Diff) are going to be unreviewable - but the remaining commits should be commit-by-commit.

Fixes https://github.com/hashicorp/pandora/issues/3583


mbfrahry (approved):
LGTM

PR #3742, opened 2024-02-01
`tools/generator-go-sdk` and `tools/generator-terraform`: porting `containerservice` over to `hashicorp/go-azure-sdk`s base layer

tombuildsstuff:
This PR will update `ContainerService` to use the new base layer.

Since this Service contains generated resources - this is going to break the Terraform Generation temporarily - so we'll need to manually update `hashicorp/terraform-provider-azurerm` to use the updated version of `hashicorp/go-azure-sdk` and patch the generated resources - at which point the automation can be re-triggered.

This PR intentionally only flips `ContainerService` over to the new base layer and doesn't remove the old base layer logic from `generator-terraform` (although this is no longer used) to give us a rollback path should there be some issue.


jackofallops (approved):
LGTM 👍

PR #3744, opened 2024-02-01
`tools/generator-terraform`: removing support for the old base layer

tombuildsstuff:
This PR removes support for using SDK Clients using the old base layer from `./tools/generator-terraform` - since once #3742 has been merged, all (existing) generated resources will be exclusively using the new base layer. Any newly generated resources should be using the new base layer from the start at this point - so this should be fine to remove.

> **Note:** this doesn't remove support for the old base layer from `./tools/generator-go-sdk` since [this is still needed at this time](https://github.com/hashicorp/pandora/blob/356b3c415d30b6fcdac6600fa063d04695349082/tools/generator-go-sdk/main.go#L30-L69).

Dependent on #3742

### DIFF
--- a/tools/generator-go-sdk/main.go
+++ b/tools/generator-go-sdk/main.go
@@ -54,17 +54,12 @@ func main() {
 		"Automation@2020-01-13-preview",
 		"Automation@2021-06-22",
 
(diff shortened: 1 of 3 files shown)


catriona-m (approved):
LGTM!

PR #3753, opened 2024-02-02
`tools/importer-rest-api-specs`: adding a test assertion helper for the Swagger Parser

tombuildsstuff:
This PR introduces a test assertion helper for the Swagger Parser - the intention being to allow defining the object we're expecting rather than checking specific fields - which should reduce the size of (most of) these tests. It also has the benefit of centralising these assertions, which'll make the upcoming refactoring work in `importer-rest-api-specs` easier.

In this PR I've intentionally only switched over a couple of tests to using this for now, given the number of tests (~150) this wants doing piecemeal to make reviewing easier.

### DIFF
--- a/tools/data-api/internal/endpoints/home_page.go
+++ b/tools/data-api/internal/endpoints/home_page.go
@@ -2,12 +2,13 @@ package endpoints
 
 import (
 	"fmt"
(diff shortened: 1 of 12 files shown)


stephybun (approved):
LGTM 👍

Issue #3754, opened 2024-02-05
Meta: refactoring


tombuildsstuff:
Due to the rapidly shifting requirements from when Pandora was first built, there’s a decent chunk of tech debt which requires cleanup.

This issue looks to outline the different phases required to get there, which will both unify the models and terminology used in the codebase, reduce the complexity of `importer-rest-api-specs` - and allow us to roll out support for Microsoft Graph.

Whilst the phases are outlined in the next section - I want to call out the `notes` section at the end, and specifically the `updated terminology` section - with the intention of standardising the technology.

# Phases

These phases need to be done as a part of the refactor - _some_ of these need to be done sequentially, others can be done in parallel - but ultimately all are needed.

1. Preparation/lift & shift.
2. Updating `./tools/data-api` to use the new Models.
3. Building out the updated SDK for `./tools/data-api-sdk`.
4. Refactoring `./tools/data-api-differ` to use the new SDK.
5. Refactoring `./tools/generator-terraform` to use the new SDK.
6. Refactoring `./tools/generator-go-sdk` to use the new SDK.
7. Refactoring `./tools/wrapper-automation` to use the new SDK.
8. Refactoring `./tools/importer-rest-api-specs` to use the new test assertion helpers.
9. Preparation for cleanup of `./tools/importer-rest-api-specs`.
10. Refactoring `dataapigeneratorjson`.
11. Building out `./tools/data-api-repository`.
12. Refactoring `./tools/importer-rest-api-specs`.
13. Updating `./tools/importer-rest-api-specs/components/terraform` to use the new SDK types throughout.
14. Updating `./tools/importer-rest-api-specs/components/parser` to use the new SDK types.
15. Updating `./tools/importer-msgraph-metadata` so that it uses the new repository.
16. Updating `./tools/version-bumper` and `./tools/importer-rest-api-specs`.
17. Removing `./tools/sdk`.

At the end of it, the resulting directory structure of `./tools` should look _similar_ to below:

```
├── data-api
│   ├── internal
│   │   ├── commands
│   │   └── endpoints
├── data-api-differ
│   ├── internal
│   │   ├── changes
│   │   ├── commands
│   │   ├── differ
│   │   ├── log
│   │   └── views
├── data-api-repository
│   ├── internal
│   │   ├── storage
│   │   └── transforms
├── data-api-sdk
│   ├── v1
│   │   └── models
├── generator-go-sdk
│   ├── internal
│   │   ├── featureflags
│   │   └── generator
[...]

PR #3761, opened 2024-02-06
`tools/data-api-sdk`: scaffolding out the Data API SDK

tombuildsstuff:
This commit introduces the new Data API SDK such that it can be used elsewhere. At this point in time this only supports the Health endpoint - but threads through the `Available Source Data Type` concept too - which enables other tools to be updated.

This is the first 3 phases of step 1 outlined in #3754


mbfrahry (approved):
LGTM

PR #3802, opened 2024-02-15
`tools/generator-go-sdk`: refactoring prior to moving to the new Data API SDK/Models [Phase 6 Part 1]

tombuildsstuff:
This PR refactors `generator-go-sdk` by moving the existing logic into an `internal` folder - and threads through a couple of pieces needed to migrate this tool over to using the new Data API SDK/Models, namely a logging instance (which whilst configured is currently unused) and support for specifying the Source Data Type.

For the sake of consistency with the other tooling, I've updated the Generator to use a `generate` subcommand - but the logic for parsing `SourceDataType` may want lifting into the Data API SDK in a future PR.


catriona-m (approved):
LGTM!

manicminer (approved):

PR #3808, opened 2024-02-16
`tools/importer-rest-api-specs`: updating (half) the tests to use the new test assertion helpers

tombuildsstuff:
This PR updates some of the Swagger Parser tests over to using the new test helper tooling - which'll make refactoring this project over to using the new Data API SDK types easier in the future (and has the side-benefit of halving the amount of code needed per test).

Given the number of tests which need to be converted over I've opted to split this PR in two - with `constants` and `models` still to go.

Part of #3754

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/temp_removed_after_sdk_migration.go
+++ b/tools/importer-rest-api-specs/components/parser/temp_removed_after_sdk_migration.go
@@ -1,0 +1,80 @@
+package parser
+
+import (
+	"github.com/hashicorp/go-azure-helpers/lang/pointer"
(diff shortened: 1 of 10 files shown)


tools/importer-rest-api-specs/components/parser/temp_removed_after_sdk_migration.go
```diff
@@ -0,0 +1,80 @@
+package parser
```
tombuildsstuff, line 1: This file exists only temporarily/until the SDK migration is completed - it just means we can reuse these helper functions early

katbyte (approved):
LGTM 🚚

PR #3815, opened 2024-02-19
`tools/importer-rest-api-specs`: refactoring the `CommonSchema` and `Discriminator` tests to use the new test assertion helpers [Phase 8 Part 3]

tombuildsstuff:
This PR refactors both the `CommonSchema` and `Discriminator` tests to use the new test assertion helpers which'll make the upcoming models switch easier.

This also removed some unnecessary "inline" tests since Object Definitions are parsed out the same way regardless at this point, so providing we get _any_ ObjectDefinition back (which we will, and that's guaranteed post the upcoming refactor) then we'll be sure to have passed through this logic. The top-level "models" parsing (which is the last grouping of tests that need refactoring for Stage 8) will be refactored next - which is what provides the test coverage for these removed tests).


katbyte (approved):
LGTM 🎶

PR #3822, opened 2024-02-20
`tools/importer-rest-api-specs`: refactoring the remaining parser tests to use the new test assertion helpers

tombuildsstuff:
This PR refactors the remaining parser tests to use the new test assertion helpers - which both makes these easier to understand and will make switching these over to the new SDK models simpler.

A few bugs have been identified in the parser as part of updating these that I've added TODOs for - we can pull these out into issues in the not-too-distant future.


catriona-m (approved):
LGTM!

PR #3826, opened 2024-02-20
Improve unmarshalling for binary payloads

manicminer:
This removes the double pointer workaround for unmarshaling binary payloads into a `[]byte`.

Needs coordination with corresponding base layer PR: https://github.com/hashicorp/go-azure-sdk/pull/890


tools/data-api/internal/commands/serve.go
```diff
@@ -59,6 +59,10 @@ func (c ServeCommand) Run(args []string) int {
 		serviceNames = pointer.To(strings.Split(serviceNamesRaw, ","))
 	}
 
+	if logLevel := strings.TrimSpace(os.Getenv("PANDORA_LOG")); logLevel != "" {
```
tombuildsstuff, line 62: I've not made the change to the Data API yet, but the other tools are now using `LOG_LEVEL` fwiw, probably worth doing that here too:

```suggestion
	if logLevel := strings.TrimSpace(os.Getenv("LOG_LEVEL")); logLevel != "" {
```

(in the other tools this is done in `main.go` - but that's a minor nit we can clear up later if needed)

tombuildsstuff (approved):
one minor comment around `LOG_LEVEL` but this otherwise LGTM 👍

tools/data-api/internal/commands/serve.go
manicminer: Ah cool, I will align it with the rest of the tooling 👍

PR #3831, opened 2024-02-21
`tools/importer-rest-api-specs`: splitting the `dataapigeneratorjson` package into transforms and stages

tombuildsstuff:
This PR refactors the `dataapigeneratorjson` package in two ways:

1. Firstly it splits out the transformation logic (for mapping the `resourcemanager` types into the `dataapimodels` types)  into it's own package (`transforms`).
2. Secondly it updates the logic used to write the API Definitions to disk to use stages, using an intermediate `filesystem` type (similar to how the `importer-msgraph-metadata` logic works) - the intention here is to allow both testability (since we can validate the files exist at the paths we're expecting more easily) and that this type can be reused when loading the data from disk too.

This PR also makes the indentation consistent across all of the generated API Definitions - which means that when this PR is merged the regeneration is going to be fairly large but will only contain whitespace - but means that the API Definitions are consistent going forwards.

This PR represents a preparation stage for Phase 9 and 10 of https://github.com/hashicorp/pandora/issues/3754


tools/importer-rest-api-specs/components/dataapigeneratorjson/transforms/sdk_model.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("mapping fields for model %q: %+v", modelName, err)
+	}
+
+	dataApiModel := dataapimodels.Model{
+		Name:   modelName,
+		Fields: *fields,
+	}
+
+	// NOTE: `Parent` types don't get a `TypeHintIn` or `TypeHintValue`
+	// meaning that only
```
catriona-m, line 31: is this comment missing a bit 🤔 ?

tombuildsstuff: It is (it's a lift and shift) - but this is actually fixed in one of the follow up PR's so I'm gonna leave this for now

PR #3836, opened 2024-02-21
`tools/importer-rest-api-specs`: refactoring `dataapigeneratorjson` to use the new Data API SDK Types

tombuildsstuff:
This PR refactors the `importer-rest-api-specs`'s `dataapigeneratorjson` package to use the new Data API SDK types. This also includes introducing a transformation shim to map from the existing types into the Data API SDK types so that we can gradually migrate each component from using the existing types into the Data API SDK types, rather than doing this as a big bang.

Implements Phase 9 and 10 of https://github.com/hashicorp/pandora/issues/3754


catriona-m (approved):
Looks like there are some tests failing, but otherwise LGTM!

PR #3840, opened 2024-02-21
`tools/generator-go-sdk`: ServiceFabric now uses the new base layer

catriona-m:


tombuildsstuff (approved):
As mentioned offline there maybe dragons for this one but 👍

PR #3846, opened 2024-02-22
`tools/importer-rest-api-specs`: fixing a couple bugs

tombuildsstuff:
This PR fixes a couple of bugs in `importer-rest-api-specs` where both the Fields/Operations were being used interchangeably - and the Service directories were being unintentionally created in `./api-definitions` rather than `./api-definitions/{sourceDataType}`


catriona-m (approved):
LGTM!

PR #3848, opened 2024-02-22
`tools/importer-rest-api-specs`: migrating SDK Operation Options over to the new Data API SDK Types

tombuildsstuff:
This PR refactors the SDK Operation Options over to the new Data API SDK types - meaning these are now consolidated into a single type.

For the sake of convenience for now I've opted to update the old `resourcemanager` package to reference the new SDK types as the migration goes on - since this allows refactoring components piecemeal rather than needing to migrate off the entire package at a time.


tools/importer-rest-api-specs/components/parser/operations.go
```diff
@@ -5,6 +5,7 @@ package parser
 
 import (
 	"fmt"
+	"github.com/hashicorp/pandora/tools/data-api-sdk/v1/models"
```
catriona-m, line 8: imports need sorting here

katbyte (approved):
LGTM 🧨

catriona-m (approved):
LGTM!

tools/importer-rest-api-specs/components/parser/operations.go
tombuildsstuff: This is fixed in one of the later PRs fwiw

PR #3849, opened 2024-02-22
`tools/importer-rest-api-specs`: adding an import alias for the internal models package

tombuildsstuff:
I'm doing this in one swoop rather than adhoc because it makes the migration/consolidation easier


catriona-m (approved):
LGTM!

PR #3867, opened 2024-02-23
`tools/importer-rest-api-specs`: temporarily disabling the `ReadOnly` and `Sensitive` field values

tombuildsstuff:
These aren't current used but are threaded through in a few places (and should be being used) - so for now I've opted to comment these out with the TODO used as a reference for the issue (which'll be coming when the refactoring is completed)

This enables us to switch out usages of `importerModels.FieldDetails` for `SDKField` far more easily - and then re-enable this once support has been added.


mbfrahry (approved):
LGTM!

PR #3865, opened 2024-02-23
`tools/importer-rest-api-specs`: Operations now use the `SDKOperation` type from the Data API SDK

tombuildsstuff:

PR #3870, opened 2024-02-23
`tools/importer-rest-api-specs`: updating Models to come from the new Data API SDK

tombuildsstuff:
This PR also updates the Resource ID for the User Assigned Identity ID to match the updated value from https://github.com/hashicorp/pandora/pull/3865


mbfrahry (approved):
LGTM!

tombuildsstuff:
This needs `ReadOnly` threading through so that Computed-only fields are tracked correctly - WIP locally

tombuildsstuff:
Tested this locally/reviewed the pushed changed and this is fine - but as mentioned above the GHA's will fail until the regeneration is merged - so I'm going to merge this and check the regenerated data, which should bring this back to green.

PR #3874, opened 2024-02-26
disable the code generation for fleet manager

ms-henglu:
This PR removes the fleet manager from the pandora code gen configuration, I'll add it back after azurerm's next major release.


katbyte (approved):
LGTM 🧰

PR #3875, opened 2024-02-26
`tools/importer-rest-api-specs`: fixing an issue where the Resource Name would be output rather than the API Resource

tombuildsstuff:
There's also an eventual consistency issue where the Resource IDs are flapping based on the name of the UserSpecifiedSegment differing - I've reviewed https://github.com/hashicorp/pandora/pull/3860 but I don't see anything obvious in that - however given there's several PRs in flight, I think it's worth merging those and then investigating this one.


stephybun (approved):
LGTM 👍

PR #3879, opened 2024-02-26
`tools/importer-rest-api-specs`: ensuring the Example Value is updated when the Segments are updated

tombuildsstuff:
This fixes a regression where the Example Value would flip based on the order the Resource IDs were updated in


tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_inconsistent_swagger_segments.go
```diff
@@ -5,6 +5,7 @@ package dataworkarounds
 
 import (
 	"fmt"
+	"github.com/hashicorp/pandora/tools/data-api-sdk/v1/helpers"
```
jackofallops, line 8: goimports here

jackofallops (approved):
Just a super-minor imports comment, otherwise LGTM 👍

PR #3881, opened 2024-02-26
`tools/importer-rest-api-specs`: ensuring that the list operations aren't pulled out for non-GET methods

tombuildsstuff:
This PR fixes a regression where PATCH operations would be identified as a List operation


catriona-m (approved):
LGTM!

PR #3890, opened 2024-02-27
`tools/importer-rest-api-specs`: detecting `POST` List Operations and some fixes for `ReadOnly` fields

tombuildsstuff:
This PR fixes a number of bugs related to ReadOnly fields and correctly detects List Operations using a `POST` as doing so:

* `generator-go-sdk`: ReadOnly fields are output in the same manner as `Optional` fields for now - this continues the previous behaviour (and adds TODOs to fix this in the future)
* `generator-go-sdk`: ReadOnly fields don't need a `Set` Date function, so removing this
* `importer-rest-api-specs`: detecting List Operations which use `POST` in addition to `GET`
* `importer-rest-api-specs`: updating the Data Workaround for `DevCenter` to mark `DNSPrefix` as only Required (and not Optional) - since Optional is now treated as a separate/independent value.

Since this PR now identifies ReadOnly fields, of which there's quite a lot (in particular top-level fields) this will change quite a number of the API Definitions, but that's expected.

This PR _doesn't_ fix the ReadOnly related flipping, which needs further investigation - but _appears_ to be coming from references to models in different Swagger files?


manicminer (approved):
LGTM 👍

PR #3891, opened 2024-02-27
`generator-go-sdk` - replace instances of `log.Printf` with `hclog`

stephybun:


tools/generator-go-sdk/internal/cmd/generate.go
```diff
@@ -106,7 +108,7 @@ func (g GenerateCommand) Synopsis() string {
 	return "Generates a Go SDK based on the API Definitions from the Data API"
 }
 
-func (g GenerateCommand) run(ctx context.Context, input GeneratorInput) error {
+func (g GenerateCommand) run(ctx context.Context, input GeneratorInput, logger hclog.Logger) error {
```
tombuildsstuff, line 111: Whilst we can thread `logger` through every function, it's probably worth using `logger.Log` directly (e.g. calling `logger.Log.Trace("blah")` rather than `logger.Trace("blah")` below, meaning we don't need to thread `logger` through).

The main value proposition for threading `logger` through rather than using a global instance would be if we were using the nested loggers (e.g. `logger.Named("blah").Trace("blah")`) - but it might be worth remove those and update the logging to account for that so we're logging contextually relevant lines, rather than passing loggers around everywhere?

This could also allow the `logging` package to have formatting aware functions, so that rather than calling `logger.Trace(fmt.Sprintf("blah %s", value))` we could instead call `logger.Tracef("blah %s", value)`, which'd make using the logging simpler?

WDYT?

tombuildsstuff (approved):
LGTM - although I'm wondering if it'd be worth using the central instance of `logger.Log` (and/or adding wrapper functions for formatting purposes) rather than passing around the `logger` instance everywhere?

tombuildsstuff (approved):
👍

PR #3892, opened 2024-02-27
Config: Add `Microsoft.AzureStackHCI@2023-08-01-preview`

teowa:
To onboard AzureStackHCI deploymentSetting resource. Confirmed with service team that there should be no breaking change in the future.

swagger: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/azurestackhci/resource-manager/Microsoft.AzureStackHCI/preview/2023-08-01-preview/deploymentSettings.json

related docs: https://learn.microsoft.com/en-us/azure-stack/hci/deploy/deployment-azure-resource-manager-template#step-3-deploy-using-arm-template


katbyte (approved):
LGTM 🏡

config/resource-manager.hcl
```diff
@@ -595,4 +595,4 @@ service "webpubsub" {
 service "workloads" {
   name      = "Workloads"
   available = ["2023-04-01"]
-}
\ No newline at end of file
+}
```
katbyte, line 598:
```suggestion
}

```

PR #3893, opened 2024-02-27
`tools/importer-rest-api-specs`: disabling ReadOnly for now

tombuildsstuff:
This PR disables `ReadOnly` for now due to issues with the Swagger Parser not correctly identifying these when a model comes from a separate file - but hardcodes this for the User Assigned Identity resource to keep the Terraform Generator happy.

We'll come back and address this/open an issue for this in a bit.


manicminer (approved):
LGTM 👍

PR #3904, opened 2024-02-28
`tools/importer-rest-api-specs`: cleanup / scaffolding out the `internal` package

tombuildsstuff:
This PR is doing some general cleanup of `importer-rest-api-specs`, including:

1. Scaffolding out the internal package, so that we can look to migrate the different segments in in the future
2. Removing most of the Legacy SDK since this is now unused (and the remnants will be removed shortly)
3. Removes the `schema` and `segments` commands since these are now unused
4. Scaffolds out the `logging` package to match the other tools
5. Updates the Terraform Builder logic to take `models.APIResource` as input, rather than an independent set of `Constants`, `Models`, `Operations` and `ResourceIDs` - since the Terraform package will soon be working against `models.Service`

This isn't where I'd originally planned to park this PR, but given the number of changes in the next step, this felt like the most appropriate stopping point.


tools/importer-rest-api-specs/components/terraform/resources/identification.go
```diff
@@ -235,12 +234,12 @@ func modelContainsDiscriminatedTypes(model models.SDKModel, data services.Resour
 		//if topLevelObjectDefinition.Type != resourcemanager.ReferenceApiObjectDefinitionType {
 		//	continue
 		//}
-		//nestedModel, isNestedModel := data.Schema.Models[*topLevelObjectDefinition.ReferenceName]
+		//nestedModel, isNestedModel := apiResource.Schema.Models[*topLevelObjectDefinition.ReferenceName]
```
stephybun, line 237: I can't leave a comment on line 232, but since this is a refactor and a bit of a clean up is there more context on that or an open issue we can link to?

tools/importer-rest-api-specs/components/terraform/schema/build_chaos_experiment_test.go
```diff
@@ -6,18 +6,19 @@ package schema
 import (
 	"testing"
 
+	"github.com/hashicorp/pandora/tools/data-api-sdk/v1/models"
```
stephybun, line 9: import ordering

tools/importer-rest-api-specs/components/transformer/models_to_api.go
```diff
@@ -4,20 +4,15 @@
...
-	return &services.Resource{
-		Operations: resourcemanager.ApiOperationDetails{
-			Operations: schema.Operations,
-		},
-		Schema: resourcemanager.ApiSchemaDetails{
-			Constants:   schema.Constants,
-			Models:      schema.Models,
-			ResourceIds: schema.ResourceIds,
-		},
+func ApiResourceFromModelResource(schema importerModels.AzureApiResource) (*models.APIResource, error) {
```
stephybun, line 11: Not a huge deal but I had some initial confusion when looking at the filename which was models **to** api but the function defined within it is api **from** model. Would it make sense to bring those in line one way or another?

stephybun (approved):
A few minor comments/questions, otherwise LGTM 👍

tools/importer-rest-api-specs/components/terraform/schema/build_chaos_experiment_test.go
tombuildsstuff: FWIW file this gets removed in #3913 since these tests are still disabled, so I'm going to leave this for the moment.

tools/importer-rest-api-specs/components/terraform/resources/identification.go
tombuildsstuff: Good shout.

Taking a look into the history here, surprisingly it looks like [this has always been commented out](https://github.com/hashicorp/pandora/blame/5e86ea25f09456d4d9bd15659df32ec850260039/tools/importer-rest-api-specs/components/terraform/resources/identification.go#L233) - however checking this locally it appears that this should be fine to uncomment/fix this up - so I'll push a commit to do that.

tools/importer-rest-api-specs/components/transformer/models_to_api.go
tombuildsstuff: TBH the entire `transformer` directory will be being removed once the `resourcemanager` models are removed, so I'm going to suggest we leave this as-is for now, since it'll be removed shortly

PR #3908, opened 2024-02-29
`data-api` - standardise logging in Data API

stephybun:
- Instantiates a global instance of the logging like in `generator-go-sdk`
- Adds some additional debug log messaging around the discovery and processing of services in the data API

### DIFF
--- a/tools/data-api/main.go
+++ b/tools/data-api/main.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/hashicorp/pandora/tools/data-api/internal/logging"
 	"log"
 	"os"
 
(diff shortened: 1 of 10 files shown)


tools/data-api/main.go
```diff
@@ -10,14 +10,18 @@ import (
...
-	logger := hclog.New(hclog.DefaultOptions)
-	logger.SetLevel(hclog.Trace)
+	loggingOpts := hclog.DefaultOptions
+	if v := os.Getenv("LOG_LEVEL"); v != "" {
+		loggingOpts.Level = hclog.LevelFromString(v)
+	}
+	logging.Log = hclog.New(loggingOpts)
 
-	logger.Info("Data API launched")
+	logging.Log.Info("Data API launched")
```
tombuildsstuff, line 24: _nit_ this one could become (so as to match everything else)

```suggestion
	logging.Infof("Data API launched")
```

tombuildsstuff (approved):
One minor suggestion but otherwise 👍

PR #3913, opened 2024-03-01
`tools/importer-rest-api-specs`: refactoring / removing usages of _some_ of the old Terraform models

tombuildsstuff:
This PR continues on from #3904 and updates several of the old SDKs Terraform Models to use the new Data API SDK models. I've tried grouping all of the Terraform changes together and it's too much to change at once - hence splitting this up.

This also fixes some minor bugs where the incorrect Object Definition Types were being used - by changing `typealias ObjectDefinitionTypeName = string` to `typealias ObjectDefinitionTypeName string` - to ensure that using an `SDKObjectDefinitionType` where we're expecting a `TerraformSchemaObjectDefinitionType` (and visa-versa) would become a compiler error - and finally fixing any compiler errors from this.


jackofallops (approved):
LGTM 👍

PR #3919, opened 2024-03-04
`tools/importer-rest-api-specs`: refactoring the Repository/Terraform logic

tombuildsstuff:
This PR refactors the repository within `importer-rest-api-specs` to support both `RemoveService` (to remove all information about an existing Service) and `SaveService` (to persist the API Definitions). Whilst this logic existed previously, this PR consolidates this into a `Repository` type, such that we have a single interface defining how API Definitions are read/written to disk. Whilst this is a minor change, it means that we can switch this out for testing purposes as needed in the future.

Whilst we can't do this at this stage - all directories within the `./tools/importer-rest-api-specs/components/dataapigeneratorjson` directory will be moved inside an `internal` folder once this directory is split out into `./tools/data-api-repository`.

In addition this PR does some cleanup to the `terraform` directory - namely moving the `Override` and `ResourceBuildInfo` models into the `terraform` directory.

Finally, this also updates how the Terraform Resources are built to first load all of the details for the specified API Versions and then builds any Terraform Resources. This is a stepping-stone in order to enable us to refactor the `terraform` package to take `models.Service` rather than a single `models.APIResource`, but is a good parking point for this refactor.


jackofallops (approved):
LGTM 👍

PR #3921, opened 2024-03-05
update the test dependency to remove the use of deprecated fields

ms-henglu:
To resolve the comments: https://github.com/hashicorp/terraform-provider-azurerm/pull/25010#issuecomment-1977480891


stephybun (approved):
Thanks @ms-henglu LGTM 👍

PR #3986, opened 2024-03-21
`importer-rest-api-specs` - correctly update orphaned discriminated models with information from the parent

stephybun:
There are several cases in Data Factory where the orphaned discriminated models are being parsed as non-discriminated models because the discriminator information lives in the parent.

Example:
Orphaned discriminated model: 
https://github.com/Azure/azure-rest-api-specs/blob/44a81f85be1e4797fbf5e290fc6b41d48788a6ba/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/entityTypes/Credential.json#L98-L116

Parent model: 
https://github.com/Azure/azure-rest-api-specs/blob/44a81f85be1e4797fbf5e290fc6b41d48788a6ba/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/entityTypes/Credential.json#L78-L87

This change will parse orphaned models that inherit from a parent and determine the discriminator information from the parent as well as update the orphaned model with the discriminator information.

TODO:
- A test case for this will be added in a follow up PR
- We will look at calling `removeUnusedItems` after all the data for the API has been parsed to remove orphaned models that have ended up in the data - which would have allowed us to catch this issue earlier

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/models.go
+++ b/tools/importer-rest-api-specs/components/parser/models.go
@@ -459,6 +459,28 @@ func (d *SwaggerDefinition) findOrphanedDiscriminatedModels() (*internal.ParseRe
 				return nil, fmt.Errorf("appending model %q: %+v", modelName, err)
 			}
 		}
+
+		// this catches orphaned discriminated models where the discriminator information is housed in the parent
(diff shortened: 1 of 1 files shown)


tombuildsstuff (approved):
LGTM 👍

PR #3988, opened 2024-03-21
dataworkaround - `DataFactory` `2018-06-01` add data workaround for datafactory 21822

stephybun:
Added a data workaround for DataFactory

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_datafactory_28122.go
+++ b/tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_datafactory_28122.go
@@ -40,6 +40,10 @@ func (workaroundDataFactory28122) Process(apiDefinition importerModels.AzureApiD
 	managedIdentityCredential.FieldNameContainingDiscriminatedValue = pointer.To("Type")
 	managedIdentityCredential.ParentTypeName = pointer.To("Credential")
 
+	if _, ok := resource.Models["ManagedIdentityTypeProperties"]; !ok {
+		return nil, fmt.Errorf("couldn't find model ManagedIdentityTypeProperties")
+	}
+
 	managedIdentityCredential.Fields["TypeProperties"] = models.SDKField{
 		ContainsDiscriminatedValue: false,
 		JsonName:                   "typeProperties",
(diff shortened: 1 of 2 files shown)


tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_datafactory_28122.go
```diff
@@ -0,0 +1,71 @@
...
+	managedIdentityCredential.DiscriminatedValue = pointer.To("ManagedIdentity")
+	managedIdentityCredential.FieldNameContainingDiscriminatedValue = pointer.To("Type")
+	managedIdentityCredential.ParentTypeName = pointer.To("Credential")
+
+	managedIdentityCredential.Fields["TypeProperties"] = models.SDKField{
+		ContainsDiscriminatedValue: false,
+		JsonName:                   "typeProperties",
+		ObjectDefinition: models.SDKObjectDefinition{
+			Type:          models.ReferenceSDKObjectDefinitionType,
+			ReferenceName: pointer.To("ManagedIdentityTypeProperties"),
```
tombuildsstuff, line 48: _minor_ since we're referencing this model, it might be worth validating that this Model exists here? If not it'd get caught later in the process if not, but it'd be clearer here as to what's failed without digging in so deeply?

tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_datafactory_28122.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	resource.Models["ManagedIdentityCredential"] = managedIdentityCredential
+
+	// remove the following models
+	modelNames := []string{
+		"UserAssignedManagedIdentityCredential",
+		//"UserAssignedManagedIdentityTypeProperties",
+		"SystemAssignedManagedIdentityCredential",
+		//"SystemAssignedManagedIdentityTypeProperties",
```
tombuildsstuff, line 60: We'll want to remove these two as well, else we'll have orphaned models for the moment?

```suggestion
		"UserAssignedManagedIdentityTypeProperties",
		"SystemAssignedManagedIdentityCredential",
		"SystemAssignedManagedIdentityTypeProperties",
```

tombuildsstuff (approved):
A couple of minor points but 👍

tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_datafactory_28122.go
stephybun: These don't actually exist anymore by the time we apply the data workaround, so attempting to remove them results in an error.

PR #3992, opened 2024-03-21
`tools/generator-go-sdk` - add init() in id generator

catriona-m:
This add an `init()` func to resource ids which registers them to a list of known ids that can be used for recasing.

dependent on [#221](https://github.com/hashicorp/go-azure-helpers/pull/221)

### DIFF
--- a/tools/generator-go-sdk/internal/generator/templater_id_parser.go
+++ b/tools/generator-go-sdk/internal/generator/templater_id_parser.go
@@ -41,12 +41,14 @@ import (
     "strings"
 
     "github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids"
+    "github.com/hashicorp/go-azure-helpers/resourcemanager/resourceids/recaser"
 )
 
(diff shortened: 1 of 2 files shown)


tools/generator-go-sdk/internal/generator/templater_id_parser.go
```diff
@@ -83,6 +85,16 @@ type %[1]s struct {
 	return &out, nil
 }
 
+func (r resourceIdTemplater) registerId() string {
+	wordifiedName := wordifyString(r.name)
+	return fmt.Sprintf(`
+	// Adds %[1]s to a register of known ids
```
tombuildsstuff, line 91: could we remove this, else this'll be a linter warning:

```suggestion
```

(Due to the placement of this comment, linters/IDEs will highlight that the first word should be the name of the function since comments above the function are used as a string in the documentation for the package)

tombuildsstuff (approved):
One comment but otherwise 👍

PR #3997, opened 2024-03-24
config: add cognitiveservices 2023-10-01-preview

guilhem:
This version adds many options.
Should be useful for https://github.com/hashicorp/terraform-provider-azurerm/issues/23988


tombuildsstuff (approved):
LGTM - thanks for this @guilhem

PR #4008, opened 2024-03-26
`importer-rest-api-specs` - extract discriminator information for orphans from parent

stephybun:
This PR re-enables a previous fix to extract the discriminator information for orphans from the parent and also correctly sets a value for the discriminator value by using the model name if it hasn't been specified by `x-ms-discriminator-value`.

This is behaviour is currently scoped to `DataFactory` since it's one of few? services that have discriminators defined this way.

To support testing for this the `ParseSwaggerFileForTesting` helper needed to be extended to accept a service name to be able to test the scoped behaviour.

### DIFF
--- a/tools/importer-rest-api-specs/components/parser/models_discriminators_test.go
+++ b/tools/importer-rest-api-specs/components/parser/models_discriminators_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestParseDiscriminatorsTopLevel(t *testing.T) {
-	actual, err := ParseSwaggerFileForTesting(t, "model_discriminators_simple.json")
+	actual, err := ParseSwaggerFileForTesting(t, "model_discriminators_simple.json", nil)
 	if err != nil {
 		t.Fatalf("parsing: %+v", err)
 	}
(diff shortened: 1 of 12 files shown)


tools/importer-rest-api-specs/components/parser/models_discriminators_test.go
```diff
@@ -1359,3 +1359,66 @@ func TestParseDiscriminatorsOrphanedChildWithNestedModel(t *testing.T) {
 	}
 	validateParsedSwaggerResultMatches(t, expected, actual)
 }
+
+func TestParseDiscriminatorsOrphanedChildWithoutDiscriminatorValue(t *testing.T) {
+	actual, err := ParseSwaggerFileForTesting(t, "/nestedtestdata/model_discriminators_orphaned_child_without_discriminator_value.json", pointer.To("datafactory"))
```
tombuildsstuff, line 1364: Since this functionality is scoped to DataFactory, perhaps it's worth a second test with the same swagger file for `Compute` to sanity-check that we don't unintentionally enable this elsewhere - WDYT?

tombuildsstuff (approved):
A couple of minor points but otherwise LGTM 👍

PR #4030, opened 2024-04-02
`importer-rest-api-specs` - update aks test template

stephybun:
Tests with this dependency are failing due to https://github.com/hashicorp/terraform-provider-azurerm/issues/24020

### DIFF
--- a/tools/importer-rest-api-specs/components/terraform/testing/dependencies_template.go
+++ b/tools/importer-rest-api-specs/components/terraform/testing/dependencies_template.go
@@ -158,6 +158,9 @@ resource "%[1]s_kubernetes_cluster" "test" {
     node_count = 1
     vm_size    = "Standard_DS2_v2"
   }
+  upgrade_settings {
+      max_surge = "10%%"
+  }
 
   identity {
     type = "SystemAssigned"


tools/importer-rest-api-specs/components/terraform/testing/dependencies_template.go
```diff
@@ -158,6 +158,9 @@ resource "%[1]s_kubernetes_cluster" "test" {
     node_count = 1
     vm_size    = "Standard_DS2_v2"
   }
+  upgrade_settings {
+      max_surge = "10%%"
```
tombuildsstuff, line 162: since this is being output into another Sprintf function (in the TF generator) I suspect this might need another `%` here?

```suggestion
      max_surge = "10%%%"
```

tombuildsstuff (approved):
One thought but otherwise LGTM 👍

tools/importer-rest-api-specs/components/terraform/testing/dependencies_template.go
stephybun: good point, it actually needs 2 additional %

PR #4003, opened 2024-03-25
Add compute 2024-03-01

JenGoldstrich:
👋 Hello friends, the Azure Packer plugin needs the 2024-03-01 compute library for a newer version of `galleryimageversions` to solve https://github.com/hashicorp/packer-plugin-azure/issues/395

PR #4032, opened 2024-04-02
Add 2023-07-03 compute

JenGoldstrich:
👋 Hey friends

I added compute 2024-03-01 in this PR https://github.com/hashicorp/pandora/pull/4003 to the list of APIs to generate, however I misunderstood the docs I was reading and 2024-03-01 does not contain galleryimages/galleryimageversions oops

The field (NonPersistedTPM) I need for Confidential VM work is defined here https://github.com/Azure/azure-rest-api-specs/blob/165c3cab0f1301bd62b45664c943761666af6936/specification/compute/resource-manager/Microsoft.Compute/GalleryRP/stable/2023-07-03/gallery.json#L2655, so I am confident 2023-07-03 contains what I need here

Please let me know if you think I should open another PR to remove the unused 2024 compute version, we're not using it yet at least over on the Azure Packer plugin.


config/resource-manager.hcl
```diff
@@ -92,7 +92,7 @@ service "communication" {
 }
 service "compute" {
   name      = "Compute"
-  available = ["2021-07-01", "2021-11-01", "2022-03-01", "2022-03-02", "2022-03-03", "2023-03-01", "2023-04-02", "2024-03-01"]
+  available = ["2021-07-01", "2021-11-01", "2022-03-01", "2022-03-02", "2022-03-03", "2023-03-01", "2023-04-02", "2023-07-03", "2024-03-01"]
```
stephybun, line 95:
```suggestion
  available = ["2021-07-01", "2021-11-01", "2022-03-01", "2022-03-02", "2022-03-03", "2023-03-01", "2023-04-02", "2023-07-03"]
```

stephybun (commented):
Thanks for this PR @JenGoldstrich. Since the compute RPs are quite large we try not to import and process versions that aren't in use, ~so I think it makes sense to remove `2024-03-01` since it's not in use.~ 

Update: Looks like we recently upgraded the version used in the provider from `2021-11-01` to `2024-03-01`, so we could remove `2021-11-01`

stephybun (approved):
LGTM 🐢

JenGoldstrich:
Thanks @stephybun ! :D!

PR #4061, opened 2024-04-11
loadtestservice: disable terraform generation until we support encryption

mbfrahry:
With the ongoing refactor, it doesn't seem like we'll get to expanded terraform generation features for some time so I'm hoping we can disable that generation for load test service to get a much requested feature in quickly


stephybun (approved):
Thanks @mbfrahry LGTM 👍

PR #4082, opened 2024-04-22
`importer-rest-api-specs` - remove data factory workaround 28122

stephybun:
Removing this workaround since it's been resolved by https://github.com/Azure/azure-rest-api-specs/pull/28416.

Will fix #4079


jackofallops (approved):
LGTM 🐐

PR #4111, opened 2024-05-06
`config` - add `network` `2022-07-01`

stephybun:
Adding network 2022-07-01 to facilitate updating the remaining network resources in AzureRM over to `hashicorp/go-azuer-sdk`.

Also removed old versions of services that are no longer in use in the provider to slim down `go-azure-sdk`.


tombuildsstuff (approved):
LGTM 👍

Issue #4128, opened 2024-05-10
`generator-go-sdk`: should return http repsonse in `listOperationTemplate`


wuxu92:
The HTTP response should be stored in the `LatestHttpResponse` field, even if an error occurs. This means that the code lines below should be moved before the error check at L341. By doing this, the caller can check the HTTP status code instead of the error string.\

https://github.com/hashicorp/pandora/blob/a225668ee785f68fd6892df3bb359b12b2d7066a/tools/generator-go-sdk/internal/generator/templater_methods.go#L353-L356

https://github.com/hashicorp/pandora/blob/a225668ee785f68fd6892df3bb359b12b2d7066a/tools/generator-go-sdk/internal/generator/templater_methods.go#L340-L344

PR #4129, opened 2024-05-11
`generator-go-sdk`: set `LatestHttpResponse` for List method even HTTP responses error

wuxu92:
Fixes: #4128


catriona-m (approved):
Thanks @wuxu92 LGTM!

PR #4148, opened 2024-05-21
config: importing `resources` @ `2015-11-01`

tombuildsstuff:
This is a curious choice so it's worth some context.

The Key Vaults List endpoint appears to have a series of caching issues, where the cache is only appended to/removed from, rather than being updated/invalidated upon changes happening. This means that it's possible for the cache to become outdated and contain a limited subset of Resources - as in our case this is currently returning only 2 of 10 provisioned Key Vaults.

Whilst that's arguably worth a bug report to the API, checking the Azure CLI's implementation, the command `az keyvault list` performs two queries:

1. `/subscriptions/XXX/resources?$filter=resourceType%20eq%20%27Microsoft.KeyVault%2Fvaults%27&api-version=2015-11-01`
2. `/subscriptions/XXX/providers/Microsoft.KeyVault/managedHSMs?api-version=2023-07-01`

This means that the Azure CLI itself isn't using the Key Vault List endpoint and for the life of me I can't find an explanation as to the why - so in lieu of any other options, we're going to use same older API version as the Azure CLI and populate the cache in the AzureRM Provider using both the List on the Resource Type endpoint and a List on the Resources endpoint.

I'd rather not use an API version this old when a newer one is available - but the Azure CLI has opted to do so, so there must be a reason for it - I would assume there's no cache on this API Version, but I can't (easily) confirm that. So this'll have to do.


jackofallops (approved):
LGTM 👍

PR #4167, opened 2024-05-28
`tools/importer-msgraph-metadata`: refactor to generate JSON definitions using `data-api-sdk`

manicminer:
~Changes to Data API will be added in a separate PR~ - this has been implemented in #4217

Currently this should have no net effect on data definitions for Resource Manager.

Screenshot 2024-05-29 at 15 20 24


tools/importer-msgraph-metadata/components/parser/types.go
```diff
@@ -96,61 +104,136 @@ func (m *Model) IsValid() bool {
...
+		return &sdkModels.SDKObjectDefinition{
+			NestedItem:    nil,
+			ReferenceName: f.ConstantName,
+			Type:          sdkModels.ReferenceSDKObjectDefinitionType,
+		}, nil
+	}
+
 	if f.Type == nil {
-		return nil
+		// Default to strings when the type is unknown
```
tombuildsstuff, line 163: Out of interest how often is the type unknown here?

tools/importer-msgraph-metadata/components/parser/types.go
```diff
@@ -96,61 +104,136 @@ func (m *Model) IsValid() bool {
...
+			ReferenceName: f.ModelName,
+			Type:          sdkModels.ReferenceSDKObjectDefinitionType,
+		}, nil
+
 	case DataTypeArray:
 		if f.ModelName != nil {
-			if models.Found(*f.ModelName) && models[*f.ModelName].IsValid() {
-				return pointerTo(fmt.Sprintf("List<%sModel>", *f.ModelName))
+			if !models.Found(*f.ModelName) || !models[*f.ModelName].IsValid() {
+				return nil, fmt.Errorf("field type Model encountered with invalid referenced model")
```
tombuildsstuff, line 190: This is an Array of a Model:

```suggestion
				return nil, fmt.Errorf("field type Array[Model] encountered with invalid referenced model")
```

tools/importer-msgraph-metadata/components/parser/types.go
```diff
@@ -209,8 +292,29 @@ func (ft DataType) CSType() *string {
...
+	case DataTypeBool:
+		return sdkModels.BooleanSDKObjectDefinitionType
+	case DataTypeDate, DataTypeDateTime, DataTypeTime:
+		return sdkModels.DateTimeSDKObjectDefinitionType
+	case DataTypeBinary:
+		return sdkModels.RawFileSDKObjectDefinitionType
+	}
+
+	// Fall back to string where the type is not known
+	return sdkModels.StringSDKObjectDefinitionType
```
tombuildsstuff, line 313: given this, are we likely to encounter `Unknown` above?

tools/importer-msgraph-metadata/internal/pipeline/task_translate_service.go
```diff
@@ -0,0 +1,179 @@
...
+					Type: operation.RequestType.DataApiSdkObjectDefinitionType(),
+				}
+			}
+
+			var responseObject *sdkModels.SDKObjectDefinition
+			responseObjectIsCommonType := true
+
+			for _, response := range operation.Responses {
+				if response.Type != nil && *response.Type == parser.DataTypeModel && response.ModelName != nil {
+					if !models.Found(*response.ModelName) {
```
tombuildsstuff, line 94: in errors like this we should highlight _which_ model can't be found? would mean you have the information needed to investigate this without needing to trigger an import manually, when something goes wrong?

tombuildsstuff (changes requested):
hey @manicminer 

Thanks for this PR/rebasing this.

I've taken a look through and left some comments inline but on the whole this is looking pretty good - there's 2 main things:

1. I think it'd be worth introducing a `RemoveExistingAPIDefinitions()` (or similar) function to allow both the MSGraph importer and the ARM importer to clear any existing data prior to performing an import. Whilst there's _some_ benefits to threading `RemoveService` through and performing a diff against the dataset to determine what's changed - tbh it's far easier/less code/bug prone to blow away the entire dataset and assume we're working in a clean directory each time. This will necessitate updating both the Graph and ARM importers to call that function prior to importing, but that should be fine?
2. As discussed offline, can we introduce specific Stages for the `CommonTypesConstantStage` and `CommonTypesModelsStage`? Whilst these'll likely be similar to the existing `Constant` and `Models` stage, doing this means each Stage remains responsible for the files it's creating (and that the paths aren't defined outside of that), which I think is beneficial?

Other than that though this is looking pretty good, nice one 👍 

Thanks!

tools/data-api-repository/repository/transforms/sdk_field_object_definition.go
```diff
@@ -71,13 +75,30 @@ func validateObjectDefinition(input repositoryModels.ObjectDefinition, constants
...
+		_, isLocalModel := models[*input.ReferenceName]
+		var isCommonConstant, isCommonModel bool
+		if commonTypes.Constants != nil {
+			_, isCommonConstant = commonTypes.Constants[*input.ReferenceName]
 		}
-		if isConstant && isModel {
-			return fmt.Errorf("internal-error: %q was found as BOTH a Constant and a Model", *input.ReferenceName)
+		if commonTypes.Models != nil {
+			_, isCommonModel = commonTypes.Models[*input.ReferenceName]
+		}
```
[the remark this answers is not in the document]
manicminer: Nope, that will actually currently work. But if the referenced type is supposed to be a common type, then it must exist as a common type, and vice versa.

tools/importer-msgraph-metadata/components/normalize/duplicates.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+func DeDuplicate(name string) string {
+	name2 := regexp.MustCompile(`\b(\w+)\s+$1\b`).ReplaceAllString(name, "$1")
+	fmt.Println(name2)
```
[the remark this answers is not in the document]
manicminer: oops, there's no need at all to output anything here

tools/importer-msgraph-metadata/components/parser/types.go
```diff
@@ -96,61 +104,136 @@ func (m *Model) IsValid() bool {
...
+			Required:  false,
+			Sensitive: false,
+		}
+	}
+
+	return &sdkModels.SDKModel{
+		DiscriminatedValue:                    nil,
+		FieldNameContainingDiscriminatedValue: nil,
+		Fields:                                sdkFields,
+		ParentTypeName:                        nil,
```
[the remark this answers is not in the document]
manicminer: this is temporary, there are discriminated types we need to support

tools/importer-msgraph-metadata/components/parser/types.go
manicminer: Actually found quite a few occurrences when trying to answer this (and fixed them). I've changed it to a failure state so these can be better surfaced.

tools/importer-msgraph-metadata/components/parser/resourceids.go
```diff
@@ -69,6 +72,39 @@ func (r ResourceId) ID() string {
...
+				Type:              sdkModels.StaticResourceIDSegmentType,
+				Name:              segment.Value,
+			})
+		case SegmentUserValue:
+			sdkSegments = append(sdkSegments, sdkModels.ResourceIDSegment{
+				ConstantReference: nil,
+				FixedValue:        nil,
+				Type:              sdkModels.UserSpecifiedSegment,
+				Name:              *segment.Field,
+			})
```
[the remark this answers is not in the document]
manicminer: Good shout, these _should_ be well formed but I've added that to `NewResourceId()`

tombuildsstuff:
As discussed offline yesterday, I've pulled the `data-api-repository` changes out into https://github.com/hashicorp/pandora/pull/4217, so this PR is now dependent on that

tools/importer-msgraph-metadata/internal/pipeline/task_translate_service.go
```diff
@@ -0,0 +1,181 @@
...
+				}
+
+				requestObject = &sdkModels.SDKObjectDefinition{
+					ReferenceName:             operation.RequestModel,
+					ReferenceNameIsCommonType: &requestObjectIsCommonType,
+					Type:                      sdkModels.ReferenceSDKObjectDefinitionType,
+				}
+			} else if operation.RequestType != nil {
+				requestObject = &sdkModels.SDKObjectDefinition{
+					Type: operation.RequestType.DataApiSdkObjectDefinitionType(),
```
tombuildsstuff, line 87: do we need to set `ReferenceNameIsCommonType: &requestObjectIsCommonType` here?

tools/importer-msgraph-metadata/internal/pipeline/task_translate_service.go
```diff
@@ -0,0 +1,181 @@
...
+					}
+
+					responseObject = &sdkModels.SDKObjectDefinition{
+						ReferenceName:             response.ModelName,
+						ReferenceNameIsCommonType: &responseObjectIsCommonType,
+						Type:                      sdkModels.ReferenceSDKObjectDefinitionType,
+					}
+				} else if response.Type != nil {
+					responseObject = &sdkModels.SDKObjectDefinition{
+						Type: response.Type.DataApiSdkObjectDefinitionType(),
```
tombuildsstuff, line 117: do we need to set `ReferenceNameIsCommonType: &responseObjectIsCommonType` here?

tombuildsstuff (approved):
Left a few comments inline but 👍

tools/importer-msgraph-metadata/internal/pipeline/task_translate_service.go
manicminer: We don't, as these will be builtin types (`string`, `bool` etc)

tools/importer-msgraph-metadata/internal/pipeline/task_translate_service.go
manicminer: As with the request object, `response.Type` always references a builtin type

PR #4179, opened 2024-06-03
`importer-rest-api-specs` - consolidate private link service resources into one

stephybun:
Related: https://github.com/Azure/azure-rest-api-specs/pull/29303


jackofallops (approved):
Thanks @stephybun - LGTM 🦒

PR #4181, opened 2024-06-03
refactor: splitting out `data-api-repository` into it's own module

tombuildsstuff:
Spotted whilst reviewing @manicminer's Microsoft Graph PR (https://github.com/hashicorp/pandora/pull/4167) I realised that it's going to be easier if this is split out ahead of time, to differentiate what's needed/changed for Graph (so that we can ensure this is threaded through everywhere as needed).

This PR allows the three data related tools (`data-api`, `importer-msgraph-metadata` and `importer-rest-api-specs`) to use the same models and repository for interacting with the data (`importer-rest-api-specs` uses both now, the other two will be refactored in a bit)


manicminer (approved):
Thanks @tombuildsstuff, this LGTM 👍

PR #4230, opened 2024-06-21
`tools/data-api`: fixing an issue where the input arguments weren't being picked up/used

tombuildsstuff:
When testing this I was running the `data-api` from within `./tools/data-api` and totally missed that we weren't picking up non-default paths 🙃


mbfrahry (approved):
LGTM!

PR #4273, opened 2024-07-04
`tools/importer-rest-api-specs`: refactoring the `discovery` package and scaffolding the replacement pipeline

tombuildsstuff:
This PR reworks the `discovery` package, which now returns a struct of the Data Set available for a given Service/API Version combination - rather than a list of Swagger files to parse - which can instead be used to parse this data.
This enables two things:

1. The differentiation of API Definitions vs Supplementary Data (which needs to be loaded first, initially this is the common Discriminator Implementations, but could potentially contain examples in the future).
2. Gives us an easier means of handling TypeSpec in the future - since we could return an instance of `AvailableDataSetForAPIVersion` based on the compiled specifications.

Secondly this scaffolds out the updated Pipeline - at this point-in-time this isn't fully usable (since the Parser package has to be refactored first) - but once that's done this should be pretty quick/easy to flip over.

Finally, this intentionally copies and reworks the `discovery` package rather than attempting to refactor it in place, since we're making some substantial changes to the public surface area and it was just easier to have a clear divide.

I've done my best to document the `discovery` package, but once the Parser package is completed we can look to scaffold out some updated diagrams based on this.


manicminer (approved):
Thanks @tombuildsstuff, LGTM 👍

PR #4274, opened 2024-07-04
`tools/importer-rest-api-specs`: foundational refactoring of `parser`

tombuildsstuff:
This PR does some minor refactoring to the `parser` package, namely to:

1. Standardise the import alias for the SDK Models as `sdkModels`
2. Removes the `models.AzureApiResource` type in favour of `sdkModels.APIResource` (since these are _essentially_ the same at this point)
3. Switches the `parser` package to use the regular logging setup, rather than passing an instance around

There's larger changes needed, but I think this is the most sensible breaking point prior to refactoring this package apart - so I'm opting to send this now, else I suspect that the next PR is going to be hard to review


manicminer (approved):
LGTM 👍

PR #4307, opened 2024-07-15
`tools/importer-rest-api-specs`: refactoring the Parser package

tombuildsstuff:
This PR refactors the Parser package into `apidefinitions` - and removes the old logic/pipeline

Right now this _mostly_ works and requires a few fixes around Discriminators/Supplementary Data that @manicminer is going to pick up

Mostly fixes/rounds out https://github.com/hashicorp/pandora/issues/3754

PR #4331, opened 2024-07-26
`importer`: try to merge resources of tags which share the same normalize name

### LINKED ISSUE #4323: The functions Create/Update/Read/Delete and the resource reference Id aren't generated for DevCenter.AttachedNetwork
While implementing new resource Dev Center Attached Network, I found that though Pandora has generated the package about DevCenter.AttachedNetwork to [go-azure-sdk](https://github.com/hashicorp/go-azure-sdk/tree/main/resource-manager/devcenter/2023-04-01/attachednetworkconnections). But [the package](https://github.com/hashicorp/go-azure-sdk/tree/main/resource-manager/devcenter/2023-04-01/attachednetworkconnections) doesn't include the files/functions about Create/Update/Read/Delete and the resource reference Id. So please fix this issue and generate them. Thanks.

API References:
https://github.com/Azure/azure-rest-api-specs/blob/09741dc7c9e43a8f9401d782186cac723dc4af71/specification/devcenter/resource-manager/Microsoft.DevCenter/stable/2023-04-01/devcenter.json#L1214

wuxu92:
fixes #4323

there are two similar tags share the same `normalizedName` (note one tag ends with a dot(.) ) in devcenter.json, we can merge these two tags' operation into the same resource. This may seem like a swagger issue, but we can address it here since fixing swagger could take longer.

Attached NetworkConnections: https://github.com/Azure/azure-rest-api-specs/blob/09741dc7c9e43a8f9401d782186cac723dc4af71/specification/devcenter/resource-manager/Microsoft.DevCenter/stable/2023-04-01/devcenter.json#L1260C1-L1261C1
Attached NetworkConnections. :  https://github.com/Azure/azure-rest-api-specs/blob/09741dc7c9e43a8f9401d782186cac723dc4af71/specification/devcenter/resource-manager/Microsoft.DevCenter/stable/2023-04-01/devcenter.json#L628

### DIFF
--- a/tools/data-api-sdk/v1/models/api_resource.go
+++ b/tools/data-api-sdk/v1/models/api_resource.go
@@ -3,6 +3,8 @@
 
 package models
 
+import "fmt"
+
 // APIResource defines a grouping of related information within an APIVersion.
 type APIResource struct {
 	// Constants specifies a map of Constant Name (key) to SDKConstant (value)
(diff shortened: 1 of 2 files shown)


stephybun:
Thanks for attempting to fix this @wuxu92.

We recently merged https://github.com/hashicorp/pandora/pull/4307 which consists of a refactor of the project as well as a change in approach to how we load and parse the Swagger files. As a result I believe this issue is now fixed as I see all of the available operations for `NetworkConnections` in the API definitions now.

Closing this out for now since this is resolved.

stephybun:
Thanks for bringing this to my attention @wuxu92. I need to fix some lingering side effects from the refactor to bring the project into a stable state. Once that's done I'll take a look at #4406, please bear with me.

PR #4344, opened 2024-08-05
enable container service fleet code generation

hqhqhqhqhqhqhqhqhqhqhq:
The fleet manager resource used to be auto generated like the fleet member resource. But it was using a preview api-version at beginning, the service team was asking to switch to a stable version. So the dns prefix feature was deprecated and not passed to the API anymore.


ms-henglu:
Thanks @hqhqhqhqhqhqhqhqhqhqhq for the PR, please attach the acc test results of the generated resource.

PR #4359, opened 2024-08-19
add compute@2024-07-01

ms-zhenhua:
Swagger:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2024-07-01


katbyte (approved):
LGTM 🚜

PR #4374, opened 2024-08-22
config - add network `2024-01-01-preview`

teowa:
To onboard network manager ipam pool resource:
https://github.com/Azure/azure-rest-api-specs/blob/5f24fa2aa4966a7d11892e22c0114f63ee617945/specification/network/resource-manager/Microsoft.Network/preview/2024-01-01-preview/ipAddressManager.json#L90-L91

confirmed no breaking change between this version to future stable version.

PR #4386, opened 2024-09-02
`tools/importer-rest-api-specs` - fix possible const conflict

teowa:
resolve issue in PR https://github.com/hashicorp/pandora/pull/4374
fix the issue that `strings.EquelFold` will randomly choose from capitalized or lowercased models, thus leads to conflicts in Enums, the details are in https://github.com/hashicorp/pandora/pull/4374#issuecomment-2320619349
By my local test, we can use exact match here without breaking existing generated data.


teowa:
remove `findAndMergeLocalMixins` in first time parsing specs, by test this can avoid generate two `provisioningState` consts after flatten. For existing data, this won't break any thing, and it should save some time.
https://github.com/hashicorp/pandora/blob/22811ab2048bd6af4af2814f0d462b9f60eff809/tools/importer-rest-api-specs/components/parser/flattener.go#L30-46

teowa:
superseded by #4411

Issue #4358, opened 2024-08-19
Error "'KubernetesFleetManagerResourceFleetHubProfileSchema' was not found" using generator-terraform to generate code for `kubernetes_fleet_manager` 


hqhqhqhqhqhqhqhqhqhqhq:
Trying to autogenerate code for resource `kubernetes_fleet_manager`: https://github.com/hashicorp/pandora/pull/4344

Tried running this locally following the steps:
![image](https://github.com/user-attachments/assets/7e74431d-8440-4f7c-8fd2-70df2a7b0cfb)

Below errors show up when executing step 4:
<img width="1462" alt="image" src="https://github.com/user-attachments/assets/b5b79d33-1fcb-4f35-83b5-f2b20f706b07">

Tried auto generate for older version swagger definition (2023-10-15) of the api for this resource and it worked, then failure is caused by the newly added property `hub_profile` in the latest swagger definition (2024-04-01)

manicminer:
Resolved in #4426

PR #4406, opened 2024-09-11
`importer`: fix generated terraform schema file names

wuxu92:
Fixes: #4358

The Data-api tool searches for schema files in the Terraform directory with filenames like `xxx-Resource-xxx-Schema.json`, but the filenames generated by the `importer` tool is `xxx-Resource-Schema-xxx.json`, which does not match. This PR renames the generated schema files to align with Data-api logic:

https://github.com/hashicorp/pandora/blob/6edf0c89df2b675f0acc2b9ad9e8b5bfd7fe0d03/tools/data-api-repository/repository/helpers_parse.go#L409-L414

This PR successfully generates resources for the AKS Fleet Manager:

image

image


stephybun:
Hey @wuxu92,

Resource generation is on hold for the moment since it will require significant re-work, both in terms of the architecture as well as to accommodate framework resources. As a result we're not adding support for any new generated resources and this part of the code can effectively be considered feature frozen.

I'm going to close this PR for the moment since we're currently not investing time in improvements here to support additional generated resources.

PR #4415, opened 2024-09-17
config - Adding support for AppPlatform 2024-05-01-preview to support Java_21

Chambras:
In order to support https://github.com/Azure/azure-rest-api-specs/blob/main/specification/appplatform/resource-manager/Microsoft.AppPlatform/preview/2024-05-01-preview/appplatform.json#L14130

The current implementation is using github.com/tombuildsstuff/kermit/sdk/appplatform/2023-05-01-preview/appplatform
and would like to start the work to migrate out of kermit and use hashicorp/go-azure-sdk instaed.

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -36,7 +36,7 @@ service "applicationinsights" {
 }
 service "appplatform" {
   name      = "AppPlatform"
-  available = ["2023-09-01-preview", "2023-11-01-preview", "2023-12-01", "2024-01-01-preview"]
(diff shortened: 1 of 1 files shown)


katbyte (approved):
LGTM 🎫

katbyte:
👋 hey @Chambras - it seems the generated pr from this (https://github.com/hashicorp/pandora/pull/4420) caused the change detector action to break. we have a few PRs that need to go in today so i've reverted this and reopened it as #4424 for us to dig into next week

Chambras:
Sounds good @katbyte is there anything on my side I could do to help?

PR #4421, opened 2024-09-19
importer-msgraph-metadata: improvements to reduce duplicate resource/operation names and prevent clobbering

manicminer:
- Don't trim the final word when de-duplicating names
- When building resource names, only check for verbs and OData references in the final segment of the resource ID
- Blacklist some resources (`Calendar*` and `*TermStore*`) via a service workaround, since these are bothersome and we have no need to support them
- Perform post-parsing de-duplication of operations, which can happen due to very similar URIs that are sometimes functionally equivalent anyway
  - When picking out duplicates, select the operation with the longest path and/or the longest resource ID name, as this is both deterministic and probably the most desirable outcome (longer paths tend to offer the most granular functionality)
  - Internally, store any ad-hoc request models (i.e. service-specific request models) in the operation, rather than in the global models map, as these will duplicate along with their operations and allows us to implicitly drop those at the same time

### DIFF
--- a/tools/importer-msgraph-metadata/components/parser/resourceids.go
+++ b/tools/importer-msgraph-metadata/components/parser/resourceids.go
@@ -192,7 +192,7 @@ func (r ResourceId) FullyQualifiedResourceName(suffixQualification *string) (*st
 				}
 			}
 
-			if i == len(r.Segments)+1 {
+			if i == len(r.Segments)-1 {
 				// Explicitly pluralize a $ref segment when the previous segment was a label and plural
 				if segment.Type == SegmentODataReference && segment.Value == "$ref" && r.Segments[i-1].plural {
 					segmentName = normalize.Pluralize(segmentName)
--- a/tools/importer-msgraph-metadata/components/parser/resourceids.go
+++ b/tools/importer-msgraph-metadata/components/parser/resourceids.go
@@ -63,7 +63,7 @@ func (ri ResourceIds) MatchIdOrAncestor(resourceId ResourceId) (*ResourceIdMatch
 }
 
 // ResourceId represents a unique path in Microsoft Graph that represents a resource. Resource IDs comprise a non-zero
-// number of segments, the last of which must always of type SegmentUserValue (i.e. specified by the user).
+// number of segments, the last of which must always be of type SegmentUserValue (i.e. specified by the user).
 // Whilst paths in Microsoft Graph can comprise different types of object identifiers, we currently only support simple
 // SegmentUserValue segments where the entire slug is provided. For example, these two paths are functionally equivalent:
 //
--- a/tools/importer-msgraph-metadata/components/parser/resourceids.go
+++ b/tools/importer-msgraph-metadata/components/parser/resourceids.go
@@ -86,6 +86,16 @@ func (r ResourceId) ID() string {
 	return "/" + strings.Join(segments, "/")
 }
 
+// ResourceIdName returns a name for the ResourceId. This calls FullyQualifiedResourceName with a common suffix to be
+// appended to all words preceding a user value. For example:
+//
+//	/groups/{groupId}/photos/{photoId} becomes GroupIdPhotoId
+//	/users/{userId}/messages/{messageId}/attachments/{attachmentId} becomes UserIdMessageIdAttachmentId
+func (r ResourceId) ResourceIdName() (*string, bool) {
+	name, ok := r.FullyQualifiedResourceName(pointer.To(ResourceIdSuffix))
+	return name, ok
+}
+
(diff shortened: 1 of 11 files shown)


tools/importer-msgraph-metadata/internal/pipeline/task_parse_resources.go
```diff
@@ -586,5 +584,91 @@ func (p pipelineForService) parseResources(resourceIds parser.ResourceIds, model
 		//resource.Category = normalize.DeDuplicateName(resource.Category)
 	}
 
+	// Look for resources with no operations, as well as duplicate operations. For duplicate operations drop the one with the
```
manicminer, line 587: This is the important part of this change - after parsing all resources (and their operations) for a service, we iterate them and look for operations with duplicate names. For any duplicates that are discovered, we remove them according to the code comments, and log what we did. This prevents flapping diffs because the parsing order is nondeterministic, so for each import, either of the conflicting operations may "win". This has the side effect of losing support for some operations, however as per the code comments this is probably an acceptable compromise.

tools/importer-msgraph-metadata/components/workarounds/workaround_blacklist.go
manicminer, line 1: This re-adds a small subset of the blacklisted resources of the previous `blacklist` package - it saves time and sanity attempting to parse these since they are unsupportable and unwanted anyway.

tools/importer-msgraph-metadata/components/parser/resourceids.go
```diff
@@ -62,6 +62,16 @@ func (ri ResourceIds) MatchIdOrAncestor(resourceId ResourceId) (*ResourceIdMatch
 	return nil, false
 }
 
+// ResourceId represents a unique path in Microsoft Graph that represents a resource. Resource IDs comprise a non-zero
+// number of segments, the last of which must always of type SegmentUserValue (i.e. specified by the user).
```
stephybun, line 66:
```suggestion
// number of segments, the last of which must always be of type SegmentUserValue (i.e. specified by the user).
```

tools/importer-msgraph-metadata/components/parser/resourceids.go
```diff
@@ -62,6 +62,16 @@ func (ri ResourceIds) MatchIdOrAncestor(resourceId ResourceId) (*ResourceIdMatch
...
+// Whilst paths in Microsoft Graph can comprise different types of object identifiers, we currently only support simple
+// SegmentUserValue segments where the entire slug is provided. For example, these two paths are functionally equivalent:
+//
+//	/applications/{id}/federatedIdentityCredentials/{federatedIdentityCredentialId}
+//	/applications(appId='{appId}')/federatedIdentityCredentials/{federatedIdentityCredentialName}
```
stephybun, line 71: I'm not familiar with this notation at all, but is there a `/` missing here?
```suggestion
//	/applications/(appId='{appId}')/federatedIdentityCredentials/{federatedIdentityCredentialName}
```

tools/importer-msgraph-metadata/internal/pipeline/task_parse_resources.go
```diff
@@ -303,7 +305,7 @@ func (p pipelineForService) parseResources(resourceIds parser.ResourceIds, model
 
 			// Determine prefixes to trim from the operation name
 			prefixesToTrim := []string{
-				normalize.CleanName(p.service),
+				//normalize.CleanName(p.service),
```
stephybun, line 308: Is this still supposed to be commented?

tools/importer-msgraph-metadata/components/parser/resourceids.go
```diff
@@ -360,6 +315,9 @@ const (
 	SegmentFunction       ResourceIdSegmentType = "Function"
```
stephybun, line 315: Unrelated to this PR, but I noticed all of these new segment types which I don't believe we have in the Resource Manager Importer? I'd find it helpful if we could briefly go over these in the upcoming handovers.

stephybun (approved):
Thanks @manicminer. Left a few questions, otherwise LGTM 🍎

tools/importer-msgraph-metadata/components/parser/resourceids.go
manicminer: Alas there is not... this is largely why we don't/can't currently support this. These complex segments essentially represent _two_ segments in SDK terms, and there's also the problem of them being partly user-specified and partly static, so we'd need to rework our generated ID parsers to accommodate these.

tools/importer-msgraph-metadata/components/parser/resourceids.go
manicminer: Sure thing! These are mapped to 'static' segments in the API definitions and resulting SDK, but we need to distinguish between them during parsing.

PR #4427, opened 2024-09-19
generator-go-sdk: fix unmarshalling fields that reference a model interface

manicminer:
When a model fields references an interface (i.e. a parent model), this currently fails to unmarshal. This PR fixes that by generating a temporary struct minus those such fields, unmarshalling into it, and then calling the explicit `Unmarshal{Foo}Implementation` function for that model.

Additionally, make the `Unmarshal{Foo}Implementation` functions more forgiving, so if the discriminated value field is missing, it continues to unmarshal into, and return, a `Raw{Foo}Impl` struct instead of just returning nil.

### DIFF
--- a/tools/generator-go-sdk/internal/generator/templater_models.go
+++ b/tools/generator-go-sdk/internal/generator/templater_models.go
@@ -240,8 +240,9 @@ func (c modelsTemplater) structLinesForModel(data GeneratorData, fieldNames []st
 	for _, fieldName := range fieldNames {
 		fieldDetails := c.model.Fields[fieldName]
 
-		if excludeDiscriminatedParentMembers && fieldDetails.ObjectDefinition.ReferenceName != nil {
-			if referencedModel, ok := data.models[*fieldDetails.ObjectDefinition.ReferenceName]; ok && referencedModel.IsDiscriminatedParentType() {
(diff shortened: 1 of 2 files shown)


tools/generator-go-sdk/internal/generator/templater_models.go
```diff
@@ -292,7 +234,109 @@ func (c modelsTemplater) methods(data GeneratorData) (*string, error) {
...
+
+		topLevelObject := helpers.InnerMostSDKObjectDefinition(fieldDetails.ObjectDefinition)
+		if excludeDiscriminatedParentMembers && topLevelObject.Type == models.ReferenceSDKObjectDefinitionType {
+			if referencedModel, ok := data.models[*topLevelObject.ReferenceName]; ok && referencedModel.IsDiscriminatedParentType() {
+				// Skip fields that reference a parent model (i.e. an interface rather than a struct)
+				continue
+			}
+		}
+
+		fieldTypeName := "FIXME"
```
stephybun, line 251: Is there any context on this?

stephybun (approved):
Thanks @manicminer LGTM 🌭

tools/generator-go-sdk/internal/generator/templater_models.go
manicminer: This is a holdover, we should probably be erroring here rather than outputting `FIXME` strings, I'll look to fix that up 👍

PR #4428, opened 2024-09-20
Support caller-specified RetryFuncs

manicminer:
### Changes

* Add a `Type` to `SDKOperationOption` to accommodate additional options that are orthogonal to the existing marshallable options (which become headers, query string arguments, or odata settings).
* Support a caller-specified `Content-Type` header via `SDKOperationOptionTypeContentType`, useful for file uploads and other scenarios with a customizable request body.
* Support a caller-specified `RequestRetryFunc` via `SDKOperationOptionTypeRequestFunc`, more details below.
* For MS Graph API definitions, add a `RetryFunc` option to all operations.
* For MS Graph API definitions, add a `ContentType` option whenever the request body is a binary payload.

### Background

Currently we have some hardcoded RetryFuncs to handle specific error cases where we want to either safely retry the original request, to elect to _not_ retry a request that would ordinarily be retried (e.g. 429 response), or to provide a custom error message for certain failures. This is already plumbed in and uses go-retryablehttp for backoff and retry handling.

There is already some hardcoded use of RetryFuncs, for example [in the Resource Manager Base Client](https://github.com/hashicorp/go-azure-sdk/blob/4c5d3e520faacb3faa502950a48c4db9f11e0d6e/sdk/client/resourcemanager/retry_policies.go), but individual SDKs can currently also supply a RetryFunc.

With Microsoft Graph, there are lots of edge cases where certain errors need specific handling, and it's not possible to autogenerate these - nor would it be desirable since they would apply for every invocation of an operation absent of any usage context. It's therefore necessary to expose a means of supply a RetryFunc to the caller, so they can be inlined into the AzureAD provider.

This implementation makes it possible for importers to include an optional operation option to allow the caller to specify a RetryFunc. By default this is not added, and doesn't count towards possible options, so this addition will not affect the signatures of any existing operation methods (e.g. in Resource Manager SDKs, existing or future). I've elected to have the MS Graph importer add this option to every operation.

Hamilton has its own equivalent of RetryFuncs, which since it's a handwritten SDK are built into the clients. Some examples: [service principals](https://github.com/manicminer/hamilton/blob/af82bb6bb88e2b06a1b6c1793876f73d2fe34a7b/msgraph/serviceprincipals.go#L65-L75), [users](https://github.com/manicminer/hamilton/blob/af82bb6bb88e2b06a1b6c1793876f73d2fe34a7b/msgraph/users.go#L64-L69), [groups](https://github.com/manicminer/hamilton/blob/af82bb6bb88e2b06a1b6c1793876f73d2fe34a7b/msgraph/groups.go#L64-L69), [OAuth2 grants](https://github.com/manicminer/hamilton/blob/af82bb6bb88e2b06a1b6c1793876f73d2fe34a7b/msgraph/delegated_permission_grants_client.go#L68-L77), [access package resource requests](https://github.com/manicminer/hamilton/blob/main/msgraph/accesspackageresourcerequest.go#L66-L68)... the list goes on.

### DIFF
--- a/tools/data-api-repository/repository/internal/models/operations.go
+++ b/tools/data-api-repository/repository/internal/models/operations.go
@@ -76,4 +76,8 @@ type Option struct {
 
 	// OptionsObjectDefinition describes the information contained within the Field
 	ObjectDefinition OptionObjectDefinition `json:"optionsObjectDefinition"`
+
+	// RetryFunc is true when this option specifies a client.RequestRetryFunc that can be passed in. This
+	// overrides HeaderName, ODataFieldName and QueryString when true. Has no effect when false.
+	RetryFunc bool
 }
--- a/tools/data-api-repository/repository/internal/models/operations.go
+++ b/tools/data-api-repository/repository/internal/models/operations.go
@@ -51,6 +51,12 @@ type Operation struct {
 }
 
 type Option struct {
+	// Type signals a special behavior for this Option.
+	//   Data: this option specifies Request data, as described in ObjectDefinition, HeaderName, ODataFieldName and/or QueryStringName.
+	//   ContentType: this option specifies a custom Content Type for the Request to be specified by the caller.
+	//   RetryFunc: this option specifies a client.RequestRetryFunc that can be passed in.
+	Type string
+
 	// HeaderName is the name of the Http Header which this Option should be set into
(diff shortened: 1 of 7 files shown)


stephybun (approved):
Thanks @manicminer LGTM 🍝

Issue #3325, opened 2023-11-13
Data API V2: Swagger parser contains duplicate models


stephybun:
The new data API is switching the value for the `description` field erroneously - see https://github.com/hashicorp/pandora/pull/3320/files.

This is happening in this part of the code where we find the name of the model in the parsed swagger data and return it:
https://github.com/hashicorp/pandora/blob/9bfd111d964594b9cdd39db9bc86aacf52f70fbc/tools/importer-rest-api-specs/components/parser/models.go#L362-L367

The swagger definitions contain duplicate models, some with Description populated and others without
![image (1)](https://github.com/hashicorp/pandora/assets/25562180/89bf41b9-e60d-4f70-b59a-15c7b075468a)
![image (2)](https://github.com/hashicorp/pandora/assets/25562180/7683e954-e562-43ae-943e-1f35525ec220)
![image](https://github.com/hashicorp/pandora/assets/25562180/89ca837c-92df-4faa-b38c-071dc8eed042)
![image (3)](https://github.com/hashicorp/pandora/assets/25562180/57c33d9b-7beb-4872-b7ed-b9a2ec6057c8)

Note that we did not observe this behaviour in the old data API because the `description` field wasn't exposed.

PR #4434, opened 2024-09-23
importer-rest-api-specs: temporary workaround for #3325 to avoid flapping diff in API definitions

manicminer:


tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_redis_22407.go
```diff
@@ -27,11 +27,11 @@ func (w workaroundRedis22407) Process(apiDefinition importerModels.AzureApiDefin
...
-					Required:    false,
-					ReadOnly:    false,
-					Sensitive:   false,
-					JsonName:    "notify-keyspace-events",
-					Description: "The KeySpace Events which should be monitored.",
+					Required:  false,
+					ReadOnly:  false,
+					Sensitive: false,
+					JsonName:  "notify-keyspace-events",
+					//Description: "The KeySpace Events which should be monitored.", // TODO: avoid flapping diff in api definitions, see https://github.com/hashicorp/pandora/issues/3325
```
stephybun, line 34: I don't think this needs to be commented out since it's a workaround and we're adding this data instead of attempting to parse it from the spec

stephybun (approved):
Thanks @manicminer LGTM 🥟

tools/importer-rest-api-specs/components/parser/dataworkarounds/workaround_redis_22407.go
manicminer: Cool, I've undone that 👍

PR #4437, opened 2024-09-23
importer-msgraph-metadata: workaround for roleManagement/CreateEntitlementManagementRoleAssignment method which actually returns 200

manicminer:


stephybun (approved):
Thanks @manicminer LGTM 🍕

PR #4445, opened 2024-09-24
importer-msgraph-metadata: global fix for inconsistent HTTP status codes in responses

manicminer:
Additionally, internally, collapse `Responses: []Response` into the `Operation` struct, to better model the Data API SDK types, as we are only ever using a single `Response`.


katbyte (approved):
LGTM ⛵

PR #4449, opened 2024-09-25
generator-go-sdk: allow configurable deletion of existing resources within an API version prior to generation, and enable for Microsoft Graph

manicminer:


tools/generator-go-sdk/internal/generator/helpers.go
```diff
@@ -2,11 +2,31 @@ package generator
...
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/hashicorp/pandora/tools/data-api-sdk/v1/models"
 )
 
+func CleanAndRecreateWorkingDirectory(path string) error {
+	// rm -r 💥
```
katbyte, line 12: should this be an emoji ?

katbyte (approved):
LGTM aside from 1 comment ♻️

tools/generator-go-sdk/internal/generator/helpers.go
manicminer: yep, it is a cautionary comment :firecracker:

PR #4464, opened 2024-10-04
feat: add api 2024-06-01-preview of search service

jan-mrm:
I'm interested in adding support for issue https://github.com/hashicorp/terraform-provider-azurerm/issues/26575 and for that I think the first action is to add the preview api version `2024-06-01-preview` for the search service.


stephybun (approved):
Thanks @jan-mrm LGTM 👍

PR #4465, opened 2024-10-04
`workarounds` - normalize example value in resource ID user specified segment to prevent erroneous diffs

stephybun:
At some point we began experiencing flapping diffs in the example value for resource ID segments 

Example PR of this phenomenon:
https://github.com/hashicorp/pandora/pull/4433
https://github.com/hashicorp/pandora/pull/4461

This workaround was added to standardize the resource ID segments to be consistent when resource IDs that are actually the same are defined with different user defined segments in the swagger e.g.

`/subscription/{subscriptionId}/resourceGroup/resourceGroupName/provider/Microsoft.SomeProvider/virtualMachines/{virtualMachineName}`
`/subscription/{subscriptionId}/resourceGroup/resourceGroupName/provider/Microsoft.SomeProvider/virtualMachines/{name}`

The example value which is metadata that we populate for documentation purposes when generating resources actually uses the same value that the segment name does, so this workaround should apply to the example value as well.


jackofallops (approved):
Thanks @stephybun - LGTM 🦫

PR #4468, opened 2024-10-07
config: adding DeviceRegistry 2023-11-01-preview / 2024-09-01-preview

mryanlo:
Enables device registry team to build the golang sdk for DeviceRegistry service (for the [go-azure-sdk ](https://github.com/hashicorp/go-azure-sdk/tree/main?tab=readme-ov-file) repo), then we can add DeviceRegistry to azurerm terraform: https://github.com/hashicorp/terraform-provider-azurerm

Specs to onboard: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/deviceregistry/resource-manager/Microsoft.DeviceRegistry/preview

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -198,7 +198,7 @@ service "deviceprovisioningservices" {
 }
 service "deviceregistry" {
   name      = "DeviceRegistry"
-  available = ["2023-11-01-preview", "2024-09-01-preview"]
+  available = ["2024-09-01-preview"]
 }
 service "deviceupdate" {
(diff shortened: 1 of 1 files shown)


mryanlo:
> Thanks for this PR @mryanlo. Is there a reason we need to add two separate preview versions, would the latest one suffice?

@stephybun thanks for your response. I guess the latest one should suffice for our immediate purposes. I'll just put the latest one.

stephybun (approved):
Thanks @mryanlo LGTM 👍

PR #4583, opened 2024-12-09
Add new DNS API version

alexwilcox9:
Adds support for DS and TLSA records as well as DNSSEC support

Used in https://github.com/hashicorp/terraform-provider-azurerm/pull/28213

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -222,7 +222,7 @@ service "digitaltwins" {
 }
 service "dns" {
   name      = "DNS"
-  available = ["2018-05-01"]
+  available = ["2023-07-01-preview","2018-05-01"]
 }
 service "dnsresolver" {
   name      = "DNSResolver"


stephybun (commented):
Thanks for this PR @alexwilcox9. Preview versions have a tendency to contain unexpected breaking changes, whether we allow preview versions to be used or not depends on several factors e.g. how popular/heavily used is the service, how pivotal is it in a customer's environment/infrastructure. 

`dns` is a service we would want to be really careful about not introducing unforseen breaking changes so I've asked our counterparts at Microsoft to reach out to the service team for information on whether there are plans for this to go into a stable version (in which case we would prefer to wait for that to happen instead of introducing this) and whether they can provide some assurances that this won't contain any breaking changes going forwards.

I'll let you know once we have some feedback and whether we can proceed with this.

Jingwei-MS:
Hi @alexwilcox9 , I’m Jingwei from the Azure Terraform PM team. Apologies for reaching out here, but I couldn’t find your alias. I noticed that the API you’re looking to onboard is currently in preview. Could you let me know if this preview version is expected to reach GA soon or if it’s stable enough with no anticipated breaking changes in the future? If it’s convenient, please ping me on Teams: jingweiwang, and we can have deeper discussions.

alexwilcox9:
Hi @Jingwei-MS 

I'd love to say it's about to reach GA or that it's stable but I'm afraid I don't work for Microsoft so have no access to that information

bubbletroubles:
This is also needed for https://github.com/hashicorp/terraform-provider-azurerm/issues/28732

I'll also raise it with our Microsoft reps to see if a stable version is going to be available soon.

Jingwei-MS:
Hi team, we see that DNSSEC API is still in preview and service side cannot confirm an exact ETA for GA/releasing stable APIs, and team hasn't responded to the email thread for quite long. So the feature is now pending.

Jingwei-MS:
No responses from service side, so still pending.

Jingwei-MS:
No responses yet, still pending.

PR #4594, opened 2024-12-16
importer-msgraph-metadata: workaround for `clientId` field in `oAuth2PermissionGrant` model

manicminer:
According to docs, `clientId` is not updateable so it shouldn't be specified when patching.

- https://learn.microsoft.com/en-us/graph/api/oauth2permissiongrant-update?view=graph-rest-1.0
- https://learn.microsoft.com/en-us/graph/api/oauth2permissiongrant-update?view=graph-rest-beta

Related: https://github.com/hashicorp/terraform-provider-azuread/issues/1536


stephybun (approved):
Thanks @manicminer LGTM 👍

PR #4597, opened 2024-12-17
config: adding containerservice API v2024-09-01

hqhqhqhqhqhqhqhqhqhqhq:
Add containerservice API v2024-09-01 to implement new properties such as advancedNetwork under networkProfile,

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -133,7 +133,7 @@ service "containerregistry" {
 }
 service "containerservice" {
   name      = "ContainerService"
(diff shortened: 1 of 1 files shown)


catriona-m (approved):
Thanks @hqhqhqhqhqhqhqhqhqhqhq LGTM!

jackofallops:
@hqhqhqhqhqhqhqhqhqhqhq - Can you take a look if we can remove any/all of the older versions so we can keep this relevant to the provider's needs? (current and the one we're moving to _should_ be enough)

PR #4614, opened 2025-01-09
config: adding recoveryservices API v2023-04-01 

wyattfry:


wyattfry:
In retrospect, downgrading the recoverysite api version would be too much work. it's preferable to wait until Microsoft restores the property in a future API version.

PR #4633, opened 2025-01-16
`generator-go-sdk`: add nullable support for resource manager sdk

wuxu92:
The resource manager SDK cannot support nullable values at present, so for some APIs, the JSON-MERGE-PATCH cannot be met by Pandora SDK. This PR works with https://github.com/hashicorp/go-azure-sdk/pull/1147 to add nullable support for resource manager APIs without modifying the struct definitions, which means it won't break any existing code.

Fixes #https://github.com/hashicorp/go-azure-sdk/issues/1143 also related to https://github.com/hashicorp/go-azure-sdk/issues/387

### DIFF
--- a/tools/importer-rest-api-specs/internal/components/apidefinitions/parser/dataworkarounds/workaround_nullable.go
+++ b/tools/importer-rest-api-specs/internal/components/apidefinitions/parser/dataworkarounds/workaround_nullable.go
@@ -1,0 +1,79 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package dataworkarounds
+
+import (
+	"fmt"
+
+	sdkModels "github.com/hashicorp/pandora/tools/data-api-sdk/v1/models"
+)
+
+var _ workaround = nullableWorkaround{}
+
+// this is a common workaround for nullable fields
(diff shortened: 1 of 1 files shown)


mryanlo:
Hi @wuxu92, is there any update on this PR? i am also encountering the issue where we cannot send null properties in the `go-for-azure` sdk, as said in this issue: https://github.com/hashicorp/go-azure-sdk/issues/1143

Do we have an ETA of when this PR will be merged? If this does go through, how do we consume the change?

mryanlo:
> @mryanlo Yes, currently `ObjectDefinition.Nullable` is always `false` for resource manager SDKs. To avoid significant changes to the generated SDK, I added a `nullableWorkaround` to Pandora to manually set specific targets as nullable. Although workarounds in Pandora are no longer recommended, since nullable support is only required by a limited number of resources, this approach should suffice.

@wuxu92 thanks for the response! a few more questions:
1. so we'd have to manually call `nullableWorkaround` in the pandora code for our resources? what would that look like? i see the example in the comments of the second file of this PR, but what do we do with the nullable workaround after that?
2. do we have an ETA for this PR to be merged and can be consumed?

magodo:
Instead of ad-hoc opt in the nullable behavior one by one (and regenerate the SDK), can we just update the marshal/unmarshal logic as a whole?

PR #4645, opened 2025-01-29
docs(common-ids): new Resource path links

guspan-tanadi:


stephybun (approved):
Thanks @guspan-tanadi LGTM 👍

PR #4467, opened 2024-10-07
`kubernetes_cluster_trusted_access_role_binding` - update auto-generated resource api version from `2023-03-02-preview` to `2024-05-01`

hqhqhqhqhqhqhqhqhqhqhq:
Test results for generated resource with api version `2024-05-01`
image
image

PR #4681, opened 2025-02-24
Remove `kubernetes_cluster_trusted_access_role_binding` from the auto-generated resource

ms-henglu:
to fix issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/28838

as suggested in https://github.com/hashicorp/pandora/pull/4467#issuecomment-2601830730, I'll disable the auto-gen here and change the resource to a manually written one.


stephybun (approved):
Thanks @ms-henglu LGTM 👍

PR #4692, opened 2025-02-26
dataworkaround: replace 27816 with 28837 for `dfe-string` handling change

jackofallops:


tools/importer-rest-api-specs/internal/components/apidefinitions/parser/parsingcontext/parse_object_definition.go
```diff
@@ -264,7 +264,7 @@ func (c *Context) parseDataFactoryCustomTypes(input *spec.Schema, known parserMo
 	}
 	if strings.EqualFold(formatVal, "dfe-string") {
 		return &sdkModels.SDKObjectDefinition{
-			Type: sdkModels.StringSDKObjectDefinitionType,
+			Type: sdkModels.RawObjectSDKObjectDefinitionType,
```
stephybun, line 267: I think a comment here to provide context would be good, I can see myself and others stumbling over this in the future thinking all the other mappings make sense, int -> int, bool -> bool, double -> float and then string -> interface?

stephybun (approved):
Thanks @jackofallops LGTM 🐈

PR #4697, opened 2025-03-03
config: remove `logz`

stephybun:
These resources were deprecated and removed in 4.0


jackofallops (approved):
Thanks @stephybun - LGTM 🐼

PR #4716, opened 2025-03-14
dataworkaround: add workaround for Sql Job Agent Identity 33215

sreallymatt:
Workaround for [azure-rest-api-specs#33215](https://github.com/Azure/azure-rest-api-specs/issues/33215)


mbfrahry (approved):
LGTM!

PR #4735, opened 2025-04-04
bump go version in GHA

stephybun:


.github/workflows/automation-regenerate-go-sdk.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
 
       - uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
         with:
-          go-version: '1.21.3'
+          go-version: '1.22.7'
```
jackofallops, line 24: We can leverage the .go-version file to avoid future differences:
```suggestion
          go-version-file: ./.go-version
```

jackofallops (approved):
LGTM, but for future there's an alternative:

PR #4740, opened 2025-04-07
config: Add containerservice version 2025-01-01

patst:
Adds attribute `bootstrapProfile.artifactSource` required for https://github.com/hashicorp/terraform-provider-azurerm/issues/29011


patst:
newer version now already available

PR #4767, opened 2025-05-16
add a common workaround for LROs not marked as LRO

sreallymatt:
Workaround needed to fully move `azurerm_backup_protected_vm` from `azure-sdk-for-go` to `go-azure-sdk`. Adding it as a common workaround since it looks like we'll need to use this for other resources in `recoveryservices` as well.

If this approach is acceptable, we could also remove `workaround_automation_25435.go` and replace it with `commonWorkaroundIsLRO`

### DIFF
--- a/tools/importer-rest-api-specs/internal/components/apidefinitions/parser/dataworkarounds/common_workaround_is_lro.go
+++ b/tools/importer-rest-api-specs/internal/components/apidefinitions/parser/dataworkarounds/common_workaround_is_lro.go
@@ -1,0 +1,53 @@
+package dataworkarounds
+
+import (
+	"fmt"
+	"slices"
+
+	sdkModels "github.com/hashicorp/pandora/tools/data-api-sdk/v1/models"
+)
+
+// This is a common workaround for operations that are asynchronous but don't specify the `x-ms-long-running-operation` field as `true`
+var _ workaround = commonWorkaroundIsLRO{}
+
+type commonWorkaroundIsLRO struct {
+	serviceName string
+	apiVersions []string
+	resources   []string
+	operations  []string
+}
+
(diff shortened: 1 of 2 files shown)


tools/importer-rest-api-specs/internal/components/apidefinitions/parser/dataworkarounds/common_workaround_is_lro.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		for _, operation := range w.operations {
+			o, ok := r.Operations[operation]
+			if !ok {
+				return nil, fmt.Errorf("couldn't find operation `%s`", operation)
+			}
+
+			if o.LongRunning {
+				return nil, fmt.Errorf("expected operation `%s` to not be marked as `LongRunning`. The workaround for this operation should be removed", operation)
```
mbfrahry, line 42: Just to confirm that if this runs multiple times, it won't continually error after the override has been applied?

I can't quite remember how overrides are applied and want to make sure the parser won't error if the data it's using to build the api definitions is being modified or if it's only the end product that is being modified.

mbfrahry (approved):
This looks good if my question doesn't matter!

tools/importer-rest-api-specs/internal/components/apidefinitions/parser/dataworkarounds/common_workaround_is_lro.go
sreallymatt: The only scenario where this would error is if the source data (Azure rest API spec) is updated to actually mark these operations as LROs, which is why I added this check here, so we can clean up the workaround if the API spec bug is fixed.

PR #4768, opened 2025-05-19
config: add compute version 2024-11-01

ms-zhenhua:
Reason: Support new enum value of `2022-11-01` in [networkApiVersion](https://github.com/Azure/azure-rest-api-specs/blob/760e6a6d008a3963092be9c62b8e8d193ba6bb5b/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2024-11-01/virtualMachineScaleSet.json#L5041) which is defined from [2024-11-01](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2024-11-01/virtualMachineScaleSet.json)


sreallymatt (approved):
LGTM

PR #4826, opened 2025-06-12
Container Service: Add version 2025-03-01

hqhqhqhqhqhqhqhqhqhqhq:
Tested by generating the sdk locally & passed.

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -134,7 +134,7 @@ service "containerregistry" {
 }
 service "containerservice" {
   name      = "ContainerService"
-  available = ["2019-08-01", "2023-03-02-preview", "2023-10-15", "2024-04-01", "2024-05-01", "2024-09-01", "2025-02-01", "2025-03-01"]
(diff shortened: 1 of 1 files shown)


jackofallops:
will be unblocked by https://github.com/hashicorp/go-azure-sdk/pull/1221

jackofallops (changes requested):
Thanks @hqhqhqhqhqhqhqhqhqhqhq 
Can you take a look at removing the versions that are no longer in use by the Azure RM provider here, we should be mindful of the number of versions we include and not simply add indefinitely.

Chippiewill:
Is there a reason for not adding 2025-05-01 (latest stable) instead? That version is required for node auto-provisioning.

weisdd:
I agree with @Chippiewill, let's add 2025-05-01 as well, it'll unblock https://github.com/hashicorp/terraform-provider-azurerm/issues/30190

jackofallops:
@hqhqhqhqhqhqhqhqhqhqhq - Given the feedback, is it worth skipping `2025-03-01` and jumping straight to `2025-05-01`?

PR #4871, opened 2025-07-20
feat: support policy exemptions

ma-ts:


jackofallops:
Hi @ma-ts - Thanks for this PR. We are very wary of introducing `preview` versions of APIs unless we have service guarantees that the preview features are going to be rolled forward unchanged into GA. This is a process we undertake to try and avoid breaking change in the providers. Please bare with us while we get confirmation of this.

PR #4877, opened 2025-07-23
`importer`: do not repeat plural if input is already in plural

wuxu92:
This PR is an improvement for PR #4876 as filename or tags may already in special plural format, so `skus` should not be `skuses`


jackofallops (approved):
Thanks @wuxu92 - LGTM 👍

PR #4962, opened 2025-09-17
Hanasheto/iotoperations

HanaSheto27:
Update service registration to include only 2024 version


sreallymatt (approved):
Thanks @HanaSheto27 LGTM ✅

Issue #4950, opened 2025-09-10
containerservice: regenerate to include 2025-07-01 API (AzureLinux3 OSSKU)


3mbe:
### **Summary**

The `2025-07-01` API for `Microsoft.ContainerService` adds support for `AzureLinux3` in the `osSKU` field.
Right now the generated SDK only includes up to `2025-05-01`, so `AzureLinux3` is missing.

### **Impact**

Because of this, the AzureRM provider rejects:

```hcl
os_sku = "AzureLinux3"
```

This blocks AKS clusters and node pools that require `AzureLinux3`, for example when creating clusters with Confidential VMs.

### **Request**

Please regenerate the ContainerService client with the `2025-07-01` API so that `AzureLinux3` is available in the SDK.
Once that’s in place, I’ll open a provider PR to bump the SDK and update validation.

### **References**

- [Azure REST API (2025-07-01) – Managed Clusters `osSKU`](https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update?view=rest-aks-2025-07-01&tabs=HTTP#ossku) - Shows `AzureLinux3` as a valid value.
- Open provider issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/30575
- Closed provider PR: https://github.com/hashicorp/terraform-provider-azurerm/pull/30495

PR #4965, opened 2025-09-19
add containerservice@2025-07-01

ms-henglu:
fixes https://github.com/hashicorp/pandora/issues/4950

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -134,7 +134,7 @@ service "containerregistry" {
 }
 service "containerservice" {
   name      = "ContainerService"
-  available = ["2019-08-01", "2024-04-01", "2025-02-01", "2025-03-01", "2025-05-01"]
+  available = ["2019-08-01", "2024-04-01", "2025-02-01", "2025-03-01", "2025-05-01", "2025-07-01"]
 }
 service "cosmos-db" {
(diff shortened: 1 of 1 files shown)


sreallymatt (approved):
Thanks @ms-henglu, hope you don't mind but I've rebased this on main to resolve the conflicts and get it merged. 

LGTM ✅

serge-van-der-ven:
A n00b question; in order to make use of the container services updates (e.g. osSKU AzureLinux3) I guess the Terraform azurerm provider (which we currently use) probably needs to be updated as well. Is this correct? And if so - when would that be?

sreallymatt:
Hi @serge-van-der-ven, that is correct.

The process is:

1. Pandora config gets updated with new API versions
2. Pandora will regenerate the data required for SDK generation (this data is located in `./api-definitions`)
3. Pandora will regenerate the SDK based on the API definitions that were generated from the REST API specs
4. A PR will be opened on [hashicorp/go-azure-sdk](https://github.com/hashicorp/go-azure-sdk/)
5. Once merged, that SDK will be released into a new version, at which point we can include update it in the AzureRM provider

While I can't make any guarantees (in case we run into issues, test failures, etc.), it will likely be included in next week's release.

PR #4967, opened 2025-09-19
`tooling` - Change runner to custom-linux-xl for unit tests

jackofallops:
removed fail fast as only relevant to matrix tests, and this workflow has no matrix.


catriona-m (approved):
LGTM!

PR #5009, opened 2025-10-13
add `2025-09-01` to MongoCluster

mbfrahry:


sreallymatt (commented):
@mbfrahry would you mind rebasing this on main? CI should pass now that the rest api specs submodule has been updated

PR #5013, opened 2025-10-15
add security@2023-12-01-preview

ziyeqf:
The security contact definition is only on this preview API version. This is prerquisite for issue https://github.com/hashicorp/terraform-provider-azurerm/issues/30797


ziyeqf:
Hi @jackofallops, per my understanding the resource is already GA but they are keeping use the preview api version..
And for the first question yes I will do the migration once `go-azure-sdk` ready for this resource type.

PR #5014, opened 2025-10-15
config: add mongocluster 2025-09-01

teowa:
To support CMK feature for mongo cluster resource
REST API: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/mongocluster/resource-manager/Microsoft.DocumentDB/MongoCluster/stable/2025-09-01/mongoCluster.json


teowa:
dup with https://github.com/hashicorp/pandora/pull/5009

PR #5029, opened 2025-10-24
Workaround: `Network` - remove problematic read-only property

sreallymatt:
[azure-rest-api-specs#38407](https://github.com/Azure/azure-rest-api-specs/issues/38407)


wyattfry (approved):
LGTM

PR #5092, opened 2025-12-08
Update resource-manager.hcl to add container service 2025-09-01

ms-henglu:


ms-henglu:
Hi @sreallymatt , Thank you for checking this. I don't need both, I opened this PR because https://github.com/hashicorp/pandora/pull/5091 and https://github.com/hashicorp/pandora/pull/5084 are blocked. 

(I'll close https://github.com/hashicorp/pandora/pull/5091 because it's duplicated)

PR #5103, opened 2025-12-12
config - remove `MobileNetwork` service due to retirement

sreallymatt:
Will resolve the generation errors in #5102 

Note: This service has been retired, the API specs have been removed. The Mobile Network resources will have to be removed from the `azurerm` provider outside of a major version, however since the RP has been fully removed it's not a breaking change.

`InvalidResourceNamespace: The resource namespace 'Microsoft.MobileNetwork' is invalid.`


catriona-m (approved):
Thanks @sreallymatt LGTM!

PR #5109, opened 2025-12-16
Add databricks 2026-01-01 to support new computeMode property

gerrytan:
Spec: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/databricks/resource-manager/Microsoft.Databricks/Databricks/stable/2026-01-01

This is needed to support new `computeMode` property on workspace creation payload.

submodules/rest-api-specs had to be bumped manually from `dcfc87c` to `c0f15e4` because the automated dependabot bump is not yet done and the specs were just recently merged.


gerrytan:
Hello @sreallymatt , can you please review this PR. There are demands from service team and customers to implement new Databricks feature from this version.

gerrytan:
Thank you @sreallymatt

PR #5314, opened 2026-04-09
ARM: Add ACR API 20250301-preview

magodo:
New ARM config to add ACR API 20250301-preview.

Related to: https://github.com/hashicorp/terraform-provider-azurerm/issues/32122

### DIFF
--- a/config/resource-manager.hcl
+++ b/config/resource-manager.hcl
@@ -137,7 +137,7 @@ service "containerinstance" {
 }
 service "containerregistry" {
   name      = "ContainerRegistry"
-  available = ["2019-06-01-preview", "2023-07-01", "2023-11-01-preview", "2025-04-01", "2025-11-01"]
+  available = ["2019-06-01-preview", "2023-07-01", "2023-11-01-preview", "2025-03-01-preview", "2025-04-01", "2025-11-01"]
 }
 service "containerservice" {
   name      = "ContainerService"


sreallymatt:
At a minimum, for us to consider adding another preview API version, we need answers to the questions in the PR template:

---

## Adding New API Version



> [!IMPORTANT]
> The primary purpose of API versions in this repository is to serve the **Terraform Azure Providers**. API versions are imported and maintained to support the providers' needs. If you would like to add an API version for use outside of the providers (e.g., another project consuming `go-azure-sdk`), please include a description of what you would like to use it for so we can evaluate the request.

- [ ] I have verified the Service/API Version exists in the Azure REST API Specs.
- [ ] I have followed the [Service Import Guide](../blob/main/docs/resource-manager-service-import.md).
- [ ] This PR is for another downstream project and not for the Terraform Azure Providers.





### Preview API Version



> [!IMPORTANT]
> We generally do not add preview API versions to `go-azure-sdk`, and we rarely support preview features in the provider. Exceptions require strong justification, such as when preview features have no GA date or no stable API version exists for a service. Please ensure you have thoroughly considered whether a preview API is truly necessary before proceeding.

- [ ] This PR adds a **preview** API version

1. **Is there a stable (GA) version available?**
   

2. **Why should we support this preview API?**
   

3. **Do existing resources already use a preview API?**
   

4. **What is the expected timeline for GA?**

magodo:
Hi @sreallymatt, sorry I overlooked this section in the template and removed the template completely in this PR. I add them back below:

## Adding New API Version



> [!IMPORTANT]
> The primary purpose of API versions in this repository is to serve the **Terraform Azure Providers**. API versions are imported and maintained to support the providers' needs. If you would like to add an API version for use outside of the providers (e.g., another project consuming `go-azure-sdk`), please include a description of what you would like to use it for so we can evaluate the request.

- [x] I have verified the Service/API Version exists in the Azure REST API Specs.
- [x] I have followed the [Service Import Guide](../blob/main/docs/resource-manager-service-import.md).
- [ ] This PR is for another downstream project and not for the Terraform Azure Providers.





### Preview API Version



> [!IMPORTANT]
> We generally do not add preview API versions to `go-azure-sdk`, and we rarely support preview features in the provider. Exceptions require strong justification, such as when preview features have no GA date or no stable API version exists for a service. Please ensure you have thoroughly considered whether a preview API is truly necessary before proceeding.

- [x] This PR adds a **preview** API version

1. **Is there a stable (GA) version available?**
   

    The currently active API version for the container registry task resource is 2019-06-01-preview. In the whole history, there are 4 api versions for this specific resource/api path: 2018-09-01, 2019-04-01, 2019-06-01-preview, 2025-03-01-preview. I'd assume there is no new stable api version available for this resource, unless we want to downgrade the api version to an earlier stable one.

2. **Why should we support this preview API?**
   

    There is a community feature request in the azurerm provider: https://github.com/hashicorp/terraform-provider-azurerm/issues/32122, which is asking for the new `.properties.credentials.sourceRegistry.identity` that is introduced in the 2025-03-01-preview version.

3. **Do existing resources already use a preview API?**
   

    Yes.

4. **What is the expected timeline for GA?**
   

    I haven't checked about this with the service, whilst from the model changes between 2019-06-01-preview and 2025-03-01-preview, there is only one attribute added, which is the one the users are asking for. Therefore, there shall be not compatibility concern for this version upgrade.

magodo:
@sreallymatt Could you take another look at this?

PR #5338, opened 2026-04-20
`importer`: reuse pluralize client to improve performance

wuxu92:
## Community Note

* Please vote on this PR by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original PR to help the community and maintainers prioritize for review
* Please do not leave comments along the lines of "+1", "me too" or "any updates", they generate extra noise for PR followers and do not help prioritize for review

## Description

The current pluralize function initialize the client for every call, it caused huge unnecessary memory allocations and caused GC increased a lot. by simply reusing the client, we can reduce the allocation a lot and import faster.

below is a experiment on importing Compute@2025-04-01, the allocations of pluralize client decreased from `4121064 (33.7%)` to `9841 (0.1%)` by this PR. and import time cost from 143s -> 117s.

Not reuse pluralize client:

image

Reuse pluralize client:
image

## PR Checklist

- [x] I have checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change.
- [x] I have checked if my changes close any open issues. If so, please include appropriate [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) below.
- [x] I have used a meaningful PR title to help maintainers and other users understand this change and help prevent duplicate work.


tools/importer-msgraph-metadata/components/normalize/plurals.go
```diff
@@ -27,8 +28,21 @@ var pluralExceptions = []plural{
...
 
-var singularMatchers = map[string]*regexp.Regexp{}
-var pluralMatchers = map[string]*regexp.Regexp{}
+var (
+	singularMatchers = map[string]*regexp.Regexp{}
+	pluralMatchers = map[string]*regexp.Regexp{}
+	pluralizeClient *pluralize.Client
+	once sync.Once
+	// pluralizeClient is not thread safe
+	mux sync.Mutex
```
jackofallops: I believe this is actually thread safe, do you have evidence to the contrary? (Singular() and Plural() are "read-only" and don't have any internal state change to lock for, so should be fine?)

jackofallops (changes requested):
Thanks @wuxu92 - I think we can get even more out of this by removing the mutex which I don't believe is necessary? (unless you have testing evidence to show it is?)

tools/importer-msgraph-metadata/components/normalize/plurals.go
wuxu92: Updated, Thanks! it's safe as we only call Singular and Plural here.

jackofallops (approved):

PR #5353, opened 2026-04-24
`generator-go-sdk` - expose function to programatically obtain the API version in use

jackofallops:


catriona-m (approved):
Thanks @jackofallops LGTM!

PR #5427, opened 2026-06-02
`generator-go-sdk` - remove targets no longer in data-api data.

jackofallops:
This PR addresses 2 issues:
- API versions removed from config were previously not removed from the generated versions.
- restructuring of configured versions was not correctly represented (resulting in orphaned code not being cleaned up and mis-representing the state of the API)


catriona-m (approved):
Thanks @jackofallops LGTM!

